PR #39, opened 2021-05-06
fix(DATAHUB-223): redirect to existing project after delete

dnsos:
This PR fixes a bug that occurred when deleting a project that is currently the first in a user's project list (as displayed in the sidebar). The previous behaviour was that the user was redirected to a page with an invalid ID (because the project has just been deleted).

Now we filter the `projectsForSidebar` to not include the deleted project and redirect to the first project of the `remainingProjects` on project deletion.


vogelino (approved):
Good catch! 🤝

PR #42, opened 2021-05-25
fix: remove range so all the projects are shown on the homepage

lucasoeth:


vogelino (changes requested):
Thanks to git old versions of this code remain in the git history and also, the documentation covers how to do pagination. Therefore it is better to remove the dead code entirely as it pollutes the codebase.

vogelino (approved):

PR #45, opened 2021-05-25
DATAHUB-243 & -244 - New landing and projects pages 

vogelino:
This PR creates basic page scaffolds for the discussed pages **landing** and **projects** and adds the hero as designed by @dnsos to the homepage (so it's not empty).

Neither the will the full landing page design be implemented nor will the fancy projects preview card. This is just about setting up the right base for further development.


lucasoeth (approved):
sieht super aus, kann kaum warten es auf der Webseite zu sehen

PR #46, opened 2021-05-27
DATAHUB-218 — Documentation infrastructure for Markdown content

vogelino:
This PR adds the ability to add some routes for documentation using [mdx (Markdown React)](https://github.com/mdx-js/mdx) as a content formatting language.

**In a nutshell**:
- To add a page, duplicate one of the existing pages in `/pages/docs` and adapt it to your needs (use markdown for formatting the content)
- In `/src/components/DocsSidebar/index.tsx`, add pages and/or groups of pages to the sidebar on the left
- A table of contents is automatically generated. It only uses headline of the level 2 for the sake of simplicity
- Responsive styles, as well as navigation logic (toggling sidebar on mobile), is taken care of


dnsos (approved):
Nice work @vogelino! This will add a lot of value once we have actual content for the documentation.

I haven't looked into _all_ details, but it seems like a great, stable implementation. Also, using the UI is smooth and adding content via MDX will be very helpful.

One minor thing: I think the active style of menu and TOC items was introduced here (bold blue). Because the font weight changes, there is a visual "glitch" when changing pages. Maybe a different highlighting could be used here. But this is really not crucial to this PR and can be discussed/solved separately.

vogelino:
> Because the font weight changes, there is a visual "glitch" when changing pages.

Right. This is a problem indeed. I will merge this PR and solve the problem in a separate branch. Thank you for your review!

PR #48, opened 2021-05-31
fix(ActiveLink): Fix styles of active links to avoid layout shift

vogelino:
This PR changes the style of active links, which were previously bold and could provoke layout shifts.

New look:
image


dnsos (approved):
Yes, I like that solution! 👍

PR #49, opened 2021-05-31
feat: improved token creation

dnsos:
These changes enable the user to refresh a project token. More specifically:

- a generic `TokenDisplay` component is added
- the component is used on the `account/projects/{id}/edit` route for refreshing the token
- the component is now also used when a project is newly created
- an unrelated fix in the `UserProjectsNav` component was also made


vogelino (approved):
Great work as always! 👌   

I took the liberty to fix an unrelated styling error that hasn't been introduced with your changes.

PR #51, opened 2021-06-01
Feat/256 - Project cards

vogelino:
This PR adds a map as background to the ProjectPreview cards and automatically geocodes the map based on the location string. It doesn't yet use the records' geolocation, which would be more accurate, but this will be adapted in the future.

This PR must be reviewed locally, as the vercel preview URL has authorization issues. Here is how a ProjectPreview looks like:

Screen Shot 2021-06-01 at 12 12 37 PM


src/components/ProjectPreview/index.tsx
```diff
@@ -12,38 +14,66 @@ export const ProjectPreview: FC<PublicProject> = ({
...
-        "relative overflow-hidden",
+        "relative overflow-hidden group",
       ].join(" ")}
       style={{ paddingBottom: 100 }}
     >
-      <Link href={`${id}`}>
-        <a sx={{ textDecoration: "none", color: "text" }}>
+      <Link href={`/${id}`}>
+        <a href={`/${id}`} ref={mapWrapperRef}>
+          {isInViewport && (
```
dnsos, line 46: Is there any way that we can cache what has already been in the viewport? Right now the map and the Mapbox tiles are fetched every time something comes **back** into the viewport.

Or is this on purpose because rendering so much WebGL at the same time is expensive?

dnsos (commented):
Wow, great work @vogelino! This adds such a visual quality. I also like the implementation.

One thing that concerns me is that rendering so much WebGL might cause us performance issues in the future. I see you already used `isInViewport` to probably mitigate this. Perhaps another solution is to only allow rendering a certain amount of project previews at the same time...

src/components/ProjectPreview/index.tsx
vogelino: Thanks @dnsos !

Only maps visible in the viewport are rendered and the tiles are actually loaded from the cache. I didn't notice too many problems performance-wise but there are certainly some other improvements. I'm not sure about limiting the number of projects though as it could result in a page with only some of the projects having a background map, which wouldn't look good I think. 

About the maps coming back, I agree. When we use reactQuery we can say to fetch only once and then never again. I didn't achieve this using useSWR. However, I think the results are fetched from the cache anyway but we might want to investigate further. 

Maybe we should merge this, so people can test it and then create a new ticket if performance problems persist? What do you think?

vogelino: > is this on purpose because rendering so much WebGL at the same time is expensive?

Exactly. Too many canvas on the DOM is bad for performance...

src/components/LinePath/index.tsx
```diff
@@ -11,21 +11,38 @@ import { DateValueType, LineGraphType } from "../../common/interfaces";
...
+    value: 0,
+  },
+];
+
+const normalizeData = (data: DateValueType[]): DateValueType[] =>
+  data.length <= 1 ? defaultArr : data;
+
 export const LinePath: FC<LineGraphType> = ({ width, height, data }) => {
+  const normalizedData = normalizeData(data);
+  // console.log(data, normalizedData);
```
lucasoeth, line 31: I don't know if that was left on purpose but probably better if we remove it

src/lib/hooks/usePublicProjects/index.ts
```diff
@@ -69,7 +63,7 @@ export const getPublicProjects = async (
       foreignTable: "devices.records",
       ascending: false,
     })
-    .limit(1, { foreignTable: "devices" })
+    .limit(10, { foreignTable: "devices" })
```
lucasoeth, line 66: why return 10 devices if parseDeviceRecords `return mappedDevices[0]` only one anyways.

src/components/ProjectPreview/index.tsx
dnsos: Thanks for the explanation. And I absolutely agree, let's merge this and observe the performance and how people perceive the maps! 🙂

lucasoeth (approved):
I think the code looks amazing the results look even better! Great job :) I left a few comments but it's for you to decide how important they are. I'm also gonna make a ticket, because we should discuss how we want to parse the data for the frontend that we all agree on because I feel like at the moment we all have our own bandades lol

dnsos (approved):

src/lib/hooks/usePublicProjects/index.ts
vogelino: We cannot be sure that the first device has data and we want to show the fancy area chart on top of the ProjectPreview card. That's why we grab the 10 first devices and check which one has the most records.

vogelino:
Thanks for your review @lucasoeth 

> I feel like at the moment we all have our own bandades lol

I'm open to reconsidering how we do things but I'm not sure I know what you mean. Do you have an example?

PR #52, opened 2021-06-01
Fix/257 - Better feedback for empty projects

vogelino:
This PR adds:
- A default marker based on the location when there are no records with geolocation
- A better feedback message for the chart area when there are no devices

**Screenshot:**
Screen Shot 2021-06-01 at 1 41 58 PM


dnsos (approved):
Nice work @vogelino! 👏 

Just a side note: I feel that the GitHub UI mistakenly displays that this PR changes 27 files, however most of these changes are already on `main` if I see this correctly.

dnsos:
Ah, just remembered that we decided to remove the subline "Die Autor:in ..."

vogelino:
> Nice work @vogelino! 👏
> 
> Just a side note: I feel that the GitHub UI mistakenly displays that this PR changes 27 files, however, most of these changes are already on `main` if I see this correctly.

Right. Lots of this is because I merge a branch from an already merge PR into this branch before it was reviewed so I could reuse the geocoding code.

PR #53, opened 2021-06-01
DATAHUB-100/tooltip

lucasoeth:
- add a line where cursor is
- show y value as tooltip
- show date as tooltip
- add padding above line for better looks


dnsos:
Hey @vogelino and @lucasoeth, this PR is currently throwing an error in the tests for the `useProjectTokens` hook, even though it wasn't touched. I've encountered this once before and can't quite figure out what's going on there. Do you have an idea?

lucasoeth:
> Hey @vogelino and @lucasoeth, this PR is currently throwing an error in the tests for the `useProjectTokens` hook, even though it wasn't touched. I've encountered this once before and can't quite figure out what's going on there. Do you have an idea?

Hey @dnsos, I've had this error quite a bit as well. It is reproducible locally but from my testing there is only about a 50% chance of this particular test failing.  I think we should open a ticket because it is def a bug. Sometimes it passes, sometimes it doesn't, and it is frustrating. My suspicion is that it has to do with the msw and mocking. 

If otherwise @dnsos and @vogelino think the PR is good, I would merge it.

vogelino (changes requested):
Great work overall! The tooltip looks and feels great. Very nice 💪 

I have a few comments though:
- There seems to still be the bug we noticed together. To reproduce it, deactivate mocking and navigate to the following page: http://localhost:3000/27
Screen Shot 2021-06-02 at 12 04 14 AM

- (optional) I wished the file `LineChart/index.tsx` would be split into chucks each dedicated to a specific purpose. Right now the file is pretty big and will probably keep growing.
- The storybook story for `LineChart` is broken. Please make sure it is fixed before merging. 
- I added some comments with some minimal change requests. Feel free to ignore them if you lack time.

vogelino:
> > Hey @vogelino and @lucasoeth, this PR is currently throwing an error in the tests for the `useProjectTokens` hook, even though it wasn't touched. I've encountered this once before and can't quite figure out what's going on there. Do you have an idea?
> 
> Hey @dnsos, I've had this error quite a bit as well. It is reproducible locally but from my testing there is only about a 50% chance of this particular test failing. I think we should open a ticket because it is def a bug. Sometimes it passes, sometimes it doesn't, and it is frustrating. My suspicion is that it has to do with the msw and mocking.
> 
> If otherwise @dnsos and @vogelino think the PR is good, I would merge it.

I have had this issue too and I'm quite annoyed by it too. I've been kind of lazy to fix it but we should definitively do that soon! 😄  I'll have a look tomorrow!

package.json
```diff
@@ -12,10 +12,12 @@
     "@turf/turf": "6.3.0",
     "@visx/axis": "1.7.0",
     "@visx/curve": "1.7.0",
+    "@visx/event": "^1.7.0",
```
dnsos, line 15: It's a good idea to pin dependencies already when adding them. Can be achieved with setting `save-exact = true` in your NPM config.

dnsos (commented):
Very nice work, @lucasoeth. This feature is an important addition and I like the implementation.

I also think that the `LineChart` component could be split into more pieces, e.g. separating all the tooltip stuff away from it.

Only commenting here, instead of approving, because of the open issues that @vogelino mentioned. After that we're good to go!

vogelino (approved):

dnsos (approved):
👍

PR #55, opened 2021-06-02
feat(ProjectPreview): Show the sensors amount on each ProjectPreview

vogelino:
This PR adds an indication of the number of sensors on each Project Card. Getting the aggregated sum of all devices via the supabase sdk without loading all the devices isn't possible. This is only doable via the Backend. Therefore, for now, the device limit has been removed from the query. In real-case scenarios, this is probably not a problem. I already created a backend ticket to create the sum aggregation on the backend side.


lucasoeth (approved):
Alles sieht super aus! Danke

PR #59, opened 2021-06-02
Feat/264 - Landing page "About" section

vogelino:
This PR adds the "About" section of the landing page, like in the designs.


src/components/LandingLabAbout/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+          ].join(" ")}
+        >
+          <strong className='font-bold'>226</strong> Open Source Projekte
+          <span className='block text-gray-500'>and counting</span>
+        </div>
+        <div className='xl:col-span-2 grid xl:grid-cols-2 gap-4 xl:gap-8'>
+          {furtherProjects.map(({ name, repo, logo }) => (
+            <a
+              key={name}
+              href={`https://github.com/technologiestiftung/`}
```
dnsos, line 84: Here we wanna link to the actual `repo`, right?

src/components/LandingLabAbout/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+              title={`Das Code-Repository von `}
```
dnsos, line 85: Add the `name` here

src/components/LandingProjectAbout/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                technologiestiftung/
+                <span className='text-black'>{repo}</span>
+              </a>
+              <ul>
+                {technologies.map(({ link, description, name }) => (
+                  <li key={name}>
+                    <a
+                      href={link}
+                      title={description}
+                      className='text-gray-500 underline hover:bg-blue-500 transition'
```
dnsos, line 108: The hover style looks weird with the gray text on blue background. Maybe that was actually supposed to mean `hover:text-blue-500`?

src/components/LandingLabAbout/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+        </div>
+      </div>
+      <div className='grid grid-cols-1 sm:grid-cols-2 xl:grid-cols-3 gap-4 md:gap-8 mt-12'>
+        <div
+          className={[
+            "text-2xl sm:text-3xl pb-4 pt-8 sm:pt-0 border-t",
+            "sm:border-t-0 border-gray-200",
+          ].join(" ")}
+        >
+          <strong className='font-bold'>226</strong> Open Source Projekte
```
dnsos, line 77: At some point we should fetch this number from the GitHub API, but totally fine for now as some texts are actually still placeholders anyway

dnsos (changes requested):
Amazing work, well done @vogelino!

There are just a few small things that may need to be addressed, see my comments for details.

src/components/LandingLabAbout/index.tsx
vogelino: Right! There is a bug in VS Code to be aware of: When you use the command "Wrap with an element" it removes the variables used in template strings. This is what happened here. Good catch!

src/components/LandingLabAbout/index.tsx
vogelino: Same thing here. I will insert it back

src/components/LandingProjectAbout/index.tsx
vogelino: I will try it out! Good idea!

src/components/LandingLabAbout/index.tsx
vogelino: I agree! I will look into how to do that!

vogelino:
> There are just a few small things that may need to be addressed, see my comments for details.

Thank you @dnsos for your good review! I fixed the template literals and added a hook to fetch the number of repos dynamically of the homepage about section. Check it out! :)

dnsos (approved):
Amazing, now let's merge this!

Great that you already implemented the dynamic number of repos btw 🚀

PR #60, opened 2021-06-02
feat(DATAHUB-238): enhanced linechart

dnsos:
This PR improves the `LineChart` in several ways. It

- adds horizontal grid lines for better sense-making of the graph
- displays axis labels (if they are provided)
- actually displays the labels in the `Projects` component/view
- updates the tooltip to account for more "extreme" data
- does some general housekeeping stuff

Screenshot 2021-06-02 at 21 38 49

---

The `LineChart` component is getting rather long, perhaps this should be tackled now as well?


vogelino (approved):
Great changes! Well done! Love it!

Just one thing. It seems as the value "0" on the Y-axis isn't displayed.

Screen Shot 2021-06-03 at 8 58 01 AM

julizet (changes requested):
some small adjutments within the index.tsx:

`case "Druck": return Pa` should be changed to `case "Luftdruck": return hPa` to be more specific and use the correct unit.

`case "PAXCounter": return Personen` should be changed to `case "PAXCounter": return Anzahl an Personen` to be even more precise

src/common/interfaces.ts
```diff
@@ -203,6 +203,8 @@ export interface LineGraphType {
   width: number;
   height: number;
   data: Array<DateValueType>;
+  yAxisUnit?: string;
+  xAxisUnit?: string;
```
lucasoeth, line 207: isn't the xAxis unit always datetime?

lucasoeth (changes requested):
Great changes, the chart keeps looking better! thanks

A change I would request is to add the unit into the tooltip, what are your thoughts on that @vogelino ?

src/common/interfaces.ts
dnsos: Yep, that's true. However I would leave the name of the unit flexible as a `string`, so that we're not limiting ourselves to one particular wording such as "Messdatum". Perhaps at some point we wanna rename it e.g. "Datum"...

dnsos:
> A change I would request is to add the unit into the tooltip, what are your thoughts on that @vogelino ?

Very good point, I'm gonna add it!

vogelino:
> A change I would request is to add the unit into the tooltip

I agree although we discussed different ways to do this. I would suggest having not only a string returned from your switch statement but an object containing either just the key `unit` or an additional key `tooltipFormatFunction` which is a function that accepts the value and the unit as an input and returns a string as an output. If no format function is provided, a default function combining the value with the unit (like so ``${value} ${unit}``) would be used to format the tooltip.

What do you think? This would have the advantage to be future-proof.

dnsos:
> What do you think? This would have the advantage to be future-proof.

You're right, this a very good solution. Let's future-proof this in the near future. I'm merging this for now because of the consensus in the team chat.

src/components/LineChart/index.tsx
```diff
@@ -66,7 +71,7 @@ export const LineChart = withTooltip<LineGraphType, DateValueType>(
       scale: xScale,
       values: data.map(el => el.date),
       // TODO: review how to switch to a 24 hour display
-      tickFormat: (v: Date) => timeFormat("%H:%M")(v),
+      //tickFormat: (v: Date) => timeFormat("%H:%M")(v),
```
lucasoeth, line 74: not sure if this was meant to be commented out, but might aswell remove if unused

lucasoeth (approved):
Looks good, thanks for also moving things away from theme UI to tailwind. Great work

src/components/LineChart/index.tsx
dnsos: Absolutely. It was unused because the `tickFormat` needs to be applied to the `AxisBottom` in this case, so it didn't do anything anyway.

vogelino (approved):

PR #61, opened 2021-06-03
feat(isUsernameAlreadyTaken): Verify is the username is already taken

vogelino:
This PR adds a server-side verification to the UserInfoEdit component to ensures that the username isn't already taken.


lucasoeth (approved):
Everything looks great, thanks for implementing this feature. For some reason it flew over my head when creating the user page. Solid testing as usual, learned a lot!

Maybe consider returning the only count and not the rows in the isUsernameTaken hook for smaller response size. But this is just my hyperefficient C mind talking, so please ignore hehe

vogelino:
> consider returning the only count and not the rows in the isUsernameTaken hook for smaller response size

Will do! Thx for the review! 🙏

dnsos (approved):
Looks great - very, very thorough work!

PR #63, opened 2021-06-03
feat(Accordion): Add Accordion component

vogelino:
This PR adds an Accordion component for the FAQ page:

Screen Shot 2021-06-03 at 10 47 08 PM


lucasoeth (changes requested):
Vom code her sieht alles super aus. Danke dafür.

Vom style her würde ich um drei Änderungen fragen, wenn du damit einverstanden bist:

1) If you click on one of the Acordion items, the ring effect (or however it's called) is overlapped by a lower element. It might be fixed if you give the active element a higher z-index
Screen Shot 2021-06-07 at 1 48 56 PM
2) The ability to close an Accordion item and have none open for instance
3) add some margin for better mobile compatibility

otherwise looking very fresh and feel free to merge

vogelino:
@lucasoeth Thx for your review! 👍 

As requested:
- ✅ Items can now be toggled if clicked again
- ✅ The focus ring is always in the foreground
- ✅ On mobile, there is a horizontal margin around the accordion

lucasoeth (approved):
Thanks for taking my comments into consideration. Looks good to merge now imo

PR #64, opened 2021-06-04
Chore - Remove easy peasy store

vogelino:
This PR removes the global [easy-peasy](https://flaviocopes.com/react-easy-peasy/) store as it wasn't used anymore. We might need a global store in the future, but for now, we should make sure it doesn't pollute the bundle.


dnsos (approved):
Great work @vogelino, thank you! Nice feeling being able to **remove** a dependency.

PR #65, opened 2021-06-04
Feat/272 - Landing page projects section with Map

vogelino:
This PR adds the landing page map and project slider section.


pages/index.tsx
```diff
@@ -1,16 +1,49 @@
...
 import { LandingProjectAbout } from "@components/LandingProjectAbout";
 import { LandingProjectsSlider } from "@components/LandingProjectsSlider";
-import { fakeProjects } from "@components/LandingProjectsSlider/LandingProjectsSlider.stories";
-import { FC } from "react";
+import { LandingStatsSection } from "@components/LandingStatsSection";
+import {
+  getLandingStats,
+  LandingStatsReturnType,
+} from "@lib/requests/getLandingStats";
+import { fakeCuratedProjects } from "@mocks/data";
```
dnsos, line 12: Hmm, I'm having doubts about mock data making it to production, even if we are in a very early stage. 

I would suggest defining the curated projects as a subset of our already existing mock projects - say simply the first 3. And then import a yet imaginary hook here `useCuratedProjects`. Of course we need to find a way to highlight curated projects in our production database, but for now we could just say we use the first 3 as well. What do you think @vogelino?

src/lib/requests/getLandingStats/getLandingStats.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { getLandingStats } from ".";
+
+describe("getLandingStats", () => {
+  it("should get the vewport by string", async (): Promise<void> => {
+    const stats = await getLandingStats();
+
+    expect(stats).toMatchObject({
+      usersCount: 27,
```
dnsos, line 8: Perhaps I'm missing something, but isn't this test gonna break once the actual stats change?

src/mocks/data.ts
```diff
@@ -1,5 +1,73 @@
...
+    date: "2021-04-07T12:10:01.908Z",
+    value: 0,
+  },
+  {
+    date: "2021-04-06T12:10:01.908Z",
+    value: 33,
+  },
+];
+
+export const fakeCuratedProjects = [
```
dnsos, line 36: This could be a subset of `./supabaseData/publicProjects.ts` to make it more realistic?

dnsos (changes requested):
Lovely implementation of this nice feature. Great work @vogelino!

I have a couple of comments. Most importantly I would like to discuss the use of fake projects for the slider. Even though we're at an early stage, I would suggest we take a different approach here, see my comments for ideas.

---

One responsivity note: On small screens the projects left and right are not visible, thereby keeping the user from using the slider. Probably the width of each card should be reduced so they appear slightly?

Screenshot 2021-06-10 at 09 45 58

pages/index.tsx
vogelino: Good point! I will address that. I know that this isn't ideal. But we need real data anyway. I agree that using the first three projects might be a valid solution, however, we should create them as well. Right now, the projects in our dev DB are worse than our mock data. Neither solutions seem to be great. It will probably be better when @lucasoeth has filled the v3 DB with nice-looking fake content.

src/lib/requests/getLandingStats/getLandingStats.test.tsx
vogelino: We are using mocked data here so it will only break when mock data changes, which is fine.

src/mocks/data.ts
vogelino: Good point! Will do!

pages/index.tsx
dnsos: Agree. We need realistic data in the new DB!

src/lib/requests/getLandingStats/getLandingStats.test.tsx
dnsos: Okay, got it. Somehow I hadn't realized the stats are mocked as well.

src/mocks/data.ts
vogelino: See 3539a02

pages/index.tsx
vogelino: See bd42a67

vogelino:
> On small screens the projects left and right are not visible, thereby keeping the user from using the slider. Probably the width of each card should be reduced so they appear slightly?

@dnsos  I changed it so it always shows slightly on the sides! Good suggestion!

pages/_app.tsx
```diff
@@ -15,7 +15,7 @@ import "../src/style/global.css";
 const publicURL = process.env.NEXT_PUBLIC_WEB_URL || "";
 
 if (process.env.NODE_ENV !== "production") {
-  require("../src/mocks/index");
+  // require("../src/mocks/index");
```
lucasoeth, line 18: you pushed this by accident

src/components/UserAvatar/index.tsx
```diff
@@ -26,7 +26,7 @@ export const UserAvatar: FC<UserAvatarPropType> = ({
     <BoringAvatar
       name={username}
       size={size}
-      variant='beam'
+      variant='pixel'
```
lucasoeth, line 29: glad you chose pixel :)

lucasoeth (changes requested):
Amazing Job on both the code and the design of the sliders. The code was flawless except the two things I commented about.

I also really like the slick design of it but would request two changes if you agree with them:
Firstly when clicking on one of the curated projects, when clicking the back button, you would expect to be back on the homepage but unfortunately, that is not the case. An easy fix could be for the project page to check for query parameters and if let's say 'origin=localhost:3000' is present it will replace that as the back href.
The second one is more of a personal experience and it was just that it took me a little bit of time to figure out how the slider works. I'm not sure how you can make it more obvious but maybe consider adding arrows so people know it's a slider, to begin with.

Cheers

pages/_app.tsx
vogelino: Oops! I will revert that!

vogelino:
> adding arrows so people know it's a slider

I did add arrows to help understand that is a slider

> when clicking the back button, you would expect to be back on the homepage

The back button now brings one step back in the history. Good point! 👍

lucasoeth (approved):
Thanks for taking the changes into consideration. Well implemented!

PR #67, opened 2021-06-09
Chore/move to dev3 db

lucasoeth:
In this PR I have adjusted the code to fit the new schames
changes include:

- updating the supabase types
- updating hooks referencing users table
- updating mocking for said hooks

Disclaimer: I haven't tested it to the fullest extent. After all, we got tests and I tested functionalities regarding the users account and everything works flawless. If issues should arrive they should be easily identifiable and thus I suggest merging if you guys don't find anything wrong in the changes


vogelino:
Great job! The code changes are on point! I don't see any data in the v3 dev DB and therefore when I change the credentials in my `.env` file, disable mocking and start the dev server again, I don't see any contents on the website. However, I remember you did write some SQL to generate nice-looking records, etc. Where did you use them?
I think, before using the new dev DB, we should fill it with good-looking fake data. Or am I missing something?

Cheers and great work! 👍

vogelino (approved):

PR #68, opened 2021-06-14
refactor: update typescript and hook

lucasoeth:
In this PR refactors:
- getProjects hook to more efficiently fetch data
- update typescript using the hook
- update mocking


vogelino (approved):
Great cleanings! We should go about refactoring the `Project/index.tsx` file soon but the Type improvements are already great! Well done! 👍

PR #70, opened 2021-06-30
chore(name): Change IOT Hub name to Stadtpuls everywhere but docs

vogelino:
I left the docs unchanged as @julizet is already changing them.


dnsos (commented):
Looks good to me, thank you!

Something that I don't quite understand: There are 404s now, e.g. on the homepage and on `/projects`. Is this related with your changes, @vogelino?

vogelino:
> Looks good to me, thank you!
> 
> Something that I don't quite understand: There are 404s now, e.g. on the homepage and on `/projects`. Is this related with your changes, @vogelino?

It's weird indeed. Locally this problem doesn't occur so I assume it's a vercel problem. I will redeploy to see if it helps.

vogelino:
@dnsos It way due to some environment variables not set correctly in vercel for the deploy previews. It whould work fine now (excepted for the maps)

dnsos (approved):

PR #71, opened 2021-07-07
fix(DATAHUB-274): remove fixed width from SmallModal

dnsos:
This PR removes a fixed width from the `SmallModal` that among other things is used for the sign-in form. By only setting a max-width, the Modal is now prevented from overflowing horizontally.


vogelino (approved):
Very cool! Thx for fixing that!

PR #74, opened 2021-07-13
feat(CSP): implement CSP (attempt)

dnsos:
This PR introduces a Content Security Policy to the project.

Note: This is a first attempt. Some more testing might be needed later on `main`. As we discussed earlier @vogelino


vogelino (commented):
Ok cool! Thx!

vogelino (commented):
Great! Thx for informing yourself and implementing this. One thing we should verify if matomo is still working after that as it loads an external script from our custom piwik.

PR #77, opened 2021-07-15
fix(LandingHeroImage): Url external urls for the LandingHeroImage

vogelino:
This PR cleans up the usage of base64 images in the LandingHeroImage component in an attempt to solve the 500 error appearing on the homepage


dnsos (approved):
Seems to solve the 500 error issue. This was a curious problem - good catch!

PR #79, opened 2021-07-15
feat(DATAHUB-283): Adapt logged in view according to new ci

vogelino:
This PR adapts the logged-in view to the new corporate identity


dnsos (approved):
Thanks for these updates, @vogelino. The user views look great! Couldn't find any flaws - great work!

PR #80, opened 2021-07-15
feat(DATAHUB-283): new CI on project page

dnsos:
This PR adds the new CI to the individual project page. Consider this an "MVP". We will still change the overall layout of that page, but with this PR, we're at least catching up with the new CI.

Plus: a lot of refactoring that removes Theme UI from all the components used on that page 🎉


vogelino (approved):
Looking great! Well done!

vogelino:
There are two things I noticed:
- The main headline isn't bold, when actually all other main headlines on other pages are bold
- The Blue color is mainly used for links with the exception of the table's hedline, which is also blue. Should we make it black or purple instead?

dnsos:
Hey @vogelino, thanks for the good suggestions - I made the adjustments in the last commits.

vogelino (commented):
Great changes! Thx for removing theme ui and adapting the forgotten bits to the new ci!

On the project page, I like that you used the dotted background, however, I would make sure that the whole table block and the map both have a white background (see screenshots)

Other than that, great changes! Thx!

dnsos:
> On the project page, I like that you used the dotted background, however, I would make sure that the whole table block and the map both have a white background (see screenshots)

Yes, absolutely - thanks for the pointer. It is updated now!

vogelino (approved):
Great thanks!

PR #83, opened 2021-07-20
fix(DATAHUB-305): ensure Dropdown opacity classes don't collide

dnsos:
This PR fixes the bug that currently when logged in, we cannot access the user nav because the classes `opacity-0` and `opacity-100` are simultaneously present and `opacity-0` takes precedence which results in the dropdown being invisible.

In this PR it is ensured that the classes cannot exist at the same time.


vogelino:
Great Catch! Thx for fixing! 🙏

Consider this as approved! (I'm on mobile right now)

PR #84, opened 2021-07-21
feat(matomo): Add matomo and remove CookieBanner

vogelino:
This PR adds a script to matomo like in erfrischungskarte and removes the cookie banner.


.env.example
```diff
@@ -4,3 +4,6 @@ NEXT_PUBLIC_SUPABASE_URL="https://YOUR_SUPABASE_URL.supabase.co"
 NEXT_PUBLIC_SUPABASE_PUBLIC_KEY="eyJKhbGciOisJIUzI1Nd2iIsInR5cCsI6..."
 NEXT_PUBLIC_TOKEN_API_URL="https://my-token-api.com"
 NEXT_PUBLIC_MAPBOX_TILESET_URL="mapbox://styles/mapbox/light-v10"
+NEXT_PUBLIC_MAPBOX_ACCESS_TOKEN="pk.eyVn03jxiI5HunjhdGzyw6Hszvi2dEV3WoaluGih2IZkWOBJmc1mJ9x0bnIMajVdZZiIzcGwdaJZaGO.dzF5WyVHTW5xZU4wI3j4kO"
```
dnsos, line 7: Good update! Can you perhaps remove the example on line 1 that is now obsolete?

next.config.js
```diff
@@ -29,7 +29,7 @@ module.exports = withMdxEnhanced({
               `frame-ancestors 'none'`,
               `worker-src 'self' blob:`,
               `child-src 'self' blob:`,
-              `connect-src 'self' ${process.env.NEXT_PUBLIC_TOKEN_API_URL} ${process.env.NEXT_PUBLIC_SUPABASE_URL} https://api.github.com https://api.mapbox.com https://events.mapbox.com`,
+              `connect-src 'self' ${process.env.NEXT_PUBLIC_TOKEN_API_URL} ${process.env.NEXT_PUBLIC_SUPABASE_URL}  ${process.env.NEXT_PUBLIC_MATOMO_URL} https://api.github.com https://api.mapbox.com https://events.mapbox.com`,
```
dnsos, line 32: Yes, that is indeed important to make it work. I've tried building the project locally to test if the CSP is working, but it told me that `${process.env.NEXT_PUBLIC_MATOMO_URL}` needs to also be added in the `img-src` directive of the CSP (which makes sense since we are doing image-based tracking).

src/lib/hooks/useMatomo/index.tsx
```diff
@@ -0,0 +1,53 @@
+import { useRouter } from "next/router";
+import { useEffect } from "react";
+
+const MATOMO_URL =
+  process.env.NEXT_PUBLIC_MATOMO_URL || "https://piwik.example.com";
+const MATOMO_SITE_ID = process.env.NEXT_PUBLIC_MATOMO_SITE_ID || "1";
+
+const createImageNoscript = (pathname = ""): HTMLElement => {
```
dnsos, line 8: I'm not entirely understanding how these lines make the tracking work (even though I know it works). Do you have a pointer where to find documentation about this? Or is it a completely custom implementation? If too complex to put in a sentence or two, we can also sync about this in person sometime.

dnsos (changes requested):
Thanks for implementing this! We just need to include the `NEXT_PUBLIC_MATOMO_URL` in the `img-src` directive of the CSP, then it should work.

src/lib/hooks/useMatomo/index.tsx
vogelino: It basically creates a new image each time the page changes and adds it to the DOM. The image is then loaded and is counted as a view in matomo. This is a custom implementation adapted to next.js. I'd be glad to explain in greater detail once we see each other! :)

dnsos: Ahh, I see. Yes, I'd be curious to know more about it :)

dnsos (approved):

PR #85, opened 2021-07-29
feat(STADTPULS-295): device registration tooltips

dnsos:
This PR adds tooltips to the table header of the device registration component. The columns "Device-ID" and "Anzeigename" require clarification, that's why the tooltip is added.

Screenshot 2021-08-03 at 17 48 25

Specifically, what was done is the following:

- An accessible component `QuestionMarkTooltip` was created that can be used wherever a tooltip is necessary. The component is self-built without external dependencies. It was a bit tricky to get the positioning right, but it should work now. Critical feedback welcome.
- The `QuestionMarkTooltip` was added to the `SensorsListHeaderRow` with the respective explanations.

---

Note that the initially used `reakit` dependency was removed because it was overhead and disproportionately difficult to test. A vanilla React/JS solution was built instead.


vogelino:
Great code, clean, tested, accessible, and very understandable! Very good also given the task's complexity! Well done!

I noticed two things:
- When the sensors table is empty, the tooltip is cropped (see screenshot below)
- On mobile, because we removed the header row as a quick solution, the tooltips are not reachable (see screenshot below)

Screen Shot 2021-08-04 at 3 00 52 PM
Screen Shot 2021-08-04 at 3 01 47 PM

PS: Sorry for the late review. I kinda got lost in my own work... :/

dnsos:
Thanks for the feedback, @vogelino. I have addressed your observations in the latest commits.

- The tooltip is not being cropped anymore when there are few or no rows of sensors
- That table header on mobile could indeed be improved but it seems like a separate task to me. For now I have added a single `QuestionMarkTooltip` next to the heading that explains both input rows combined. In my impression that does the job. Let me know what you think.

Screenshot 2021-08-05 at 14 33 45

vogelino (approved):
Great solution! Well done! 👏  Thx!

PR #86, opened 2021-08-03
fix(Fonts): Fix Space Grotesk in Safari

vogelino:
This PR fixes the font Space Grotesk not being displayed correctly in Safari. This was due to some bad font files being used. I downloaded the files from the official source and used it the same way they did.


dnsos (approved):
Nice! The issue was also happening in Firefox before. Now it does indeed work in both!

PR #87, opened 2021-08-03
fix(STADTPULS-268): Snap tooltip to nearest datapoint

vogelino:
This PR makes sure that line chart tooltips always snap to the nearest datapoint.


dnsos (approved):
Great! I had thought this might be trickier, but this is actually a really simple solution!

PR #88, opened 2021-08-03
fix(STADPULS-309): Buggy linechart when one record

vogelino:
This PR ensures that the line chart looks good even with one record only.

### DIFF
--- a/src/components/LinePath/index.tsx
+++ b/src/components/LinePath/index.tsx
@@ -9,37 +9,21 @@ import colors from "../../style/colors";
 const getX = (d: DateValueType): Date => new Date(d.date);
 const getY = (d: DateValueType): number => d.value;
 
-const startDate = new Date();
-const defaultArr = [
-  {
-    date: startDate.toISOString(),
-    value: 0,
-  },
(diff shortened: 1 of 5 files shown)


src/components/LinePath/index.tsx
```diff
@@ -9,37 +9,21 @@ import colors from "../../style/colors";
...
-    value: 0,
-  },
-  {
-    date: new Date(startDate.getDate() + 1).toISOString(),
-    value: 0,
-  },
-];
-
-const normalizeData = (data: DateValueType[]): DateValueType[] =>
-  data.length <= 1 ? defaultArr : data;
```
dnsos, line 25: Ahh, that's where the issue was, right?

dnsos (approved):

src/components/LinePath/index.tsx
vogelino: Right! It was initially added for the AreaChart so that there is always at least a line when no or only one record exists. The Area Chart has been ported to its own component anyway so this is unnecessary now.

PR #89, opened 2021-08-05
feat(STADTPULS-145): accept startDate & endDate in API endpoint for records

dnsos:
We can now filter the records API endpoint like so:

```
/api/v1/devices/1/records?startDate=2021-01-01T00:00:00&endDate=2021-08-06T00:00:00
```

This will be helpful when we introduce filtering by time for the users.

---

The possible query combinations (without time range | only `startDate` | only `endDate` | `startDate` and `endDate`) are handled in the `getRecordsByDeviceId` utility and the appropriate data is fetched there. `pages/api/v1/devices/[id]/records.ts` handles validation of the query params and returns status codes and hints to the user.


.vscode/settings.json
```diff
@@ -1,19 +1,19 @@
...
+    "activityBarBadge.background": "#46ECA1",
     "activityBarBadge.foreground": "#15202b",
-    "statusBar.background": "#ac31b2",
-    "statusBar.foreground": "#e7e7e7",
-    "statusBarItem.hoverBackground": "#c749cd",
-    "titleBar.activeBackground": "#ac31b2",
-    "titleBar.activeForeground": "#e7e7e7",
-    "titleBar.inactiveBackground": "#ac31b299",
-    "titleBar.inactiveForeground": "#e7e7e799"
+    "statusBar.background": "#100C53",
```
vogelino, line 10: It seems like VS Code sometimes resets the theme on its own. Not sure why. You might want to revert that.

vogelino (approved):
Really clean and super well tested! Thank you! Great initiative too as I'm working on the date time inputs now. I will be able to leverage the start/end date params when the time comes! 🤜🤛

.vscode/settings.json
dnsos: Yes, my impression is that it's related to the Peacock extension.

Anyway, this change here is actually on purpose. The theme was accidentally reset in some previous PR, here I'm changing it back to our Stadtpuls colors.

PR #90, opened 2021-08-09
feat(STADTPULS-307): ensure device ID conforms to TTN pattern

dnsos:
This PR updates the `yup` schema for device IDs. We want to make sure that inputs conform to the allowed pattern in TTN. This PR does that validation and prevents the user from inputting invalid IDs (including an update hint to what went wrong).

If the TTN pattern changes or we include other integration types, we will have to update the pattern.

We cannot rely on frontend validation only and therefore we additionally need to update our DB to only accept the new schema as well. This is yet to be done.

### DIFF
--- a/src/components/SensorsList/SensorsList.test.tsx
+++ b/src/components/SensorsList/SensorsList.test.tsx
@@ -77,7 +77,7 @@ describe("component SensorsList", () => {
     expect(idInput).toBeInTheDocument();
     expect(nameInput).toBeInTheDocument();
 
-    fireEvent.change(idInput, { target: { value: "device-B" } });
+    fireEvent.change(idInput, { target: { value: "device-b" } });
     fireEvent.change(nameInput, { target: { value: "I like to change" } });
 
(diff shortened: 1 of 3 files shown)


vogelino (approved):
Good addition! Well done!

dnsos:
@vogelino, I've added another commit (64655a3) that removes the regex check. I would argue that we should not validate something that is dictated from the outside and that can change at any time: What if suddenly `_` ore uppercase letters will be allowed in TTN and we miss that?

Therefore I would rather say to only allow more characters here. Now we also have the tooltip that clarifies that the TTN and Stadtpuls device ID have to match.

---

Sidenote: I would still wait with the merge for the [database migration](https://github.com/technologiestiftung/stadtpuls-api/pull/28), so that the allowed characters match in frontend validation and database type.

vogelino:
> I would argue that we should not validate something that is dictated from the outside and that can change at any time
I agree with that. Good decision there. Well done! 👍

PR #91, opened 2021-08-09
feat(STADTPULS-308): Improve Device Line-Chart filters

vogelino:
This PR improves the time filtering of the project page. It introduces a few components among which:
- `TimeInput`: For Hour and Minutes
- `DayPickerInput`: For Day selection
- `DatetimeRangePicker`: For start Datetime and end Datetime selection
- `DeviceLineChartFilters`: A combination of a `DatetimeRangePicker` and a few presets "24h | 7d | 30d | All"

Currently, supabase limits the amount of returned rows to 1000. We either don't know how to specify a higher limit or it isn't possible yet and we need an alternative solution (multiple calls?). However, the request headers are accurate and the code can be reviewed.


src/components/DatetimeRangePicker/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export const DatetimeRangePicker: FC = () => {
+  const [fromDate, setFromDate] = useState(new Date());
+  const [toDate, setToDate] = useState(new Date());
+  const [fromTime, setFromTime] = useState("00:00");
+  const [toTime, setToTime] = useState("00:00");
+  const modifiers = { start: fromDate, end: toDate };
+
+  return (
+    <div className='flex gap-2'>
```
dnsos, line 16: I'm wondering if we could add a `flex-wrap` here to ensure that the elements don't overflow the screen horizontally on small devices?

It would require some z-index/stack context changes though:
Screenshot 2021-08-11 at 09 48 57

dnsos (commented):
Great work! This will be a very important feature once we implement it.

The code is very clean (expecially given the complexity of the UI elements) and amazing tests as usual. 👏 

I have one comment below regarding the display on small screens. What do you think?

src/components/DatetimeRangePicker/index.tsx
vogelino: Thx for your review! Good point! I will fix that! :)

src/components/DeviceLineChartFilters/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+  );
+  const [temporalityOfRecords, setTemporaityOfRecords] = useState<TimespanType>(
+    "last24h"
+  );
+  const last7daysTimeRange = getTimeRangeByTimespan("last7days", today);
+  return (
+    <div className='border-b border-gray-100 shadow p-4 flex flex-wrap gap-8'>
+      <RadioFieldset
+        isSelected={activeFilterType === "devicesByDatetimeRange"}
+        label='Messwerte bei Zeitspanne'
```
dnsos, line 103: Hm, I would say that "Messwerte **nach** Zeitspanne" sounds more correct here

src/components/DeviceLineChartFilters/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+            last7daysTimeRange.endDateTimeString ||
+            moment().toISOString()
+          }
+          onDatetimeRangeChange={onDatetimeRangeChange}
+          tabIndex={activeFilterType === "devicesByDatetimeRange" ? 0 : -1}
+        />
+      </RadioFieldset>
+      <RadioFieldset
+        isSelected={activeFilterType === "devicesByTimespan"}
+        label='Messwerte bei Zeitraum'
```
dnsos, line 124: And here "Messwerte nach Zeitraum"

src/components/Project/index.tsx
```diff
@@ -48,15 +59,26 @@ const getCategoryUnit = (
 export const Project: FC<ProjectsType> = project => {
   const [selectedDeviceIndex, setSelectedDeviceIndex] = useState<number>(0);
   const selectedDevice = project?.devices?.[selectedDeviceIndex];
+  const [currentDatetimeRange, setCurrentDatetimeRange] = useState<{
+    startDateTimeString: string | undefined;
+    endDateTimeString: string | undefined;
+  }>({
+    startDateTimeString: today.toISOString(),
+    endDateTimeString: tenDaysAgo.toISOString(),
```
dnsos, line 67: Is it possible that the values are mixed up here? In its default state the filter seems to show the more recent date first and the `tenDaysAgo` afterwards:

Screenshot 2021-08-18 at 12 14 52

Or I am misinterpreting that?

dnsos (changes requested):
Amazing job! Seems that this task included a lot of different challenges and in my impression you solved them all very nicely. 👏 

I couldn't get into the details of _every_ bit of this PR as it is quite large, but hopefully covered the most important parts. I have made a few comments which should be relatively easy to address.

src/components/Project/index.tsx
vogelino: Good catch! I totally missed that! Thx!

dnsos (approved):
Thanks for the updates! Let's merge this!

PR #92, opened 2021-08-12
fix(STADTPULS-296): add UTC indicator

dnsos:
This is a small hotfix for STADTPULS-296. To make it understandable that we currently parse all dates/times as UTC, we should also make that clear in the UI. This way, you at least understand why e.g. your values from Berlin are 2hrs off at the moment.

This solution is not ideal, but currently we receive all records as UTC time anyway, no matter where the device is located.

### DIFF
--- a/src/components/DataTable/__snapshots__/DataTable.stories.storyshot
+++ b/src/components/DataTable/__snapshots__/DataTable.stories.storyshot
@@ -38,7 +38,7 @@ exports[`Storyshots UI Elements/DataTable Default 1`] = `
           <th
             className="py-2 px-1 font-normal text-left"
           >
-            Uhrzeit
+            Uhrzeit (UTC)
           </th>
           <th
             className="py-2 px-1 font-normal text-right"
--- a/src/components/DataTable/index.tsx
+++ b/src/components/DataTable/index.tsx
@@ -74,7 +74,7 @@ export const DataTable: React.FC<DataTableType> = ({ data, title }) => {
               <th
                 className={["py-2 px-1", "font-normal", "text-left"].join(" ")}
               >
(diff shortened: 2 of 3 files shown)


vogelino:
Oh cool! I thought it would solve my problem with supabase right now but it is only an indication. Still, very helpful! My Problem is that the server Timezone differs from the browser's timezone somehow. When I use your hook `useDeviceRecords` I get an error `time zone "t-2:00:00.000z" not recognized`. I'm not sure what I'm doing wrong and how to convert all time zones to the same format. Any idea what could be wrong?

vogelino (approved):
Good catch! A valid compromise for now! Thx!

dnsos:
> Any idea what could be wrong?

Hm, I had a similar issue, however this was solved after adding the `timestampValidator` utility. But I'm lacking a bit of context right now - maybe we can have a talk to figure out the situation?

Also in general, the issues of time and timezones are presenting us some more challenges, I'm afraid.

Thanks for the review btw! Merging this now.

vogelino:
> in general, the issues of time and timezones are presenting us some more challenges, I'm afraid.

I agree! It's a real pain! I figured out you can add the offset at the end of the ISO date string (e.g. "-02:00"). This is then correctly interpreted from supabase. I'm still not too confident about manipulating dates and it feels like a dangerous rabbithole. Anyway, it should work fine now. Thx for you help!

dnsos:
> I figured out you can add the offset at the end of the ISO date string (e.g. "-02:00"). This is then correctly interpreted from supabase.

Interesting. I'm glad it worked now. But yes, we should take a close look at this topic again in the not-so-far future.

PR #93, opened 2021-08-12
fix(STADTPULS-311): get correct lastRecordedAt

dnsos:
This PR fixes the sorting of dates in the `SensorsListWithData`.

---

- [x] fix the sorting
- [x] add a test for the function that caused the issue


dnsos:
@vogelino, feel free to close this if you've implemented the same changes in #91

vogelino (approved):
Great test! Thx for adding it! We can merge that in my opinion.

PR #95, opened 2021-08-19
fix(Project): Set last 7 days of records as default time range

vogelino:
This PR sets the last 7 days in which records were recorded as the default time range as discussed with @dnsos and @edgalindo. This ensures that, as long as a sensor recorded some data, this data will be shown and an empty chart is avoided.


dnsos (approved):
Great, thanks! Much better experience browsing through the projects/devices now.

vogelino:
> Much better experience browsing through the projects/devices now.

Thx for the quick review! 🙏

PR #97, opened 2021-08-19
fix(FormSelect): Fix select's wrong value

vogelino:
This PR ensures that selects can be both either controlled or uncontrolled while always having a value. This is important in the case an HTML form element use the form data to `onSubmit`. In this case, every input within the form must have a valid value.


dnsos (approved):
Good solution for this - it now works as expected. Thanks @vogelino

PR #99, opened 2021-09-01
reduced image size

Esshahn:
reduced filesize from 1.7mb to 600k to speed up initial load times


dnsos (approved):
Good catch @Esshahn! I've updated a snapshot that caused the test to fail (not directly related to your change). Now this should be good to go!

PR #108, opened 2021-09-08
refactor(STADTPULS-314): Refactor Emails to feel less spammy

vogelino:
This PR is not only about the files changed but about the different things done to prevent our emails to land in the recievers' SPAM folder.

**What has been done**
- [x] SPF and DMARK DNS entries have been successfully configured (And pass in all major email clients)
- [x] Email texts have been reformulated to avoid spam wordings
- [x] Added a text to ask users to check their SPAM folder too when they log in/register: _"Bitte überprüfe auch Deinen Spam-Ordner."_

**What couldn't be done**
- [ ] DKIM could not be configured correctly, because hosteurope doesn't support it (It isn't absolutely required but might improve our chance not to be classified as SPAM)
- [ ] The links within the emails have the supabase domain, which differs from the sender's email. That might cause our emails to land in SPAM. I added a CNAME record that works, but there is no way in supabase to change the domain of the confirmation link. E.g. `https://supabase.stadtpuls.com/auth/v1/verify?token=zVgRAKA2-jhsvkhUPCCF9Q&type=magiclink&redirect_to=https://stadtpuls.com/`

**I tested our emails with:**
- **Protonmail:** ✅ PASSED
- **Gmail:** ✅ PASSED
- **Yandex:** ✅ PASSED
- **Yahoo:** ✅ PASSED
- **TSB:** ✅ PASSED
- **Outlook/Hotmail:** ❌ FAILED, landed in SPAM regardless of passing SPF and DMARC policies

**Other possible reasons for outlook rejecting our email:**
- https://support.cpanel.net/hc/en-us/articles/1500007348281-Emails-go-to-spam-despite-valid-SPF-DKIM-DMARC-records-IP-Domain-Reputation
- My assumption/hope: _"Reputation/Age — How new the account is or how much email has been sent. Initially, Gmail sends all new messages to the spam folder until its reputation has improved. This is the most common reason."_

I would suggest stopping here and wait for a while to see if the emails start to go into the inbox after a while of usage. With the new DNS entries, texts, and the inquiry to check the SPAM folder, we should cover most cases. What do you think?


dnsos (commented):
Thanks for the investigation and the changes you made @vogelino! I'm not familiar with the whole SPF etc. topic, so I can't contribute much on this right now.

I only have one question: Why are we using the formal "Sie" instead of "Du" now? Is it related to the e-mail wording not being detected as spam? In general, the whole vibe of Stadtpuls feels much more like "Du". Or is there a communication shift that was discussed that I missed?

vogelino:
> Why are we using the formal "Sie" instead of "Du" now? Is it **related to the e-mail wording not being detected as spam**?

Exactly. Informal language could be interpreted as SPAM. As we here aren't advertising anything, I would consider it more important to avoid being classified as SPAM as to be exciting or in line with the way we address users. It might as well not be the reason why mails are classified as SPAM, it's hard to know at this point. Each email client uses different AI algorithms to classify their emails. Formulating the email in the most boring way, especially for login/register is an attempt to avoid future problems...

PR #109, opened 2021-09-08
Wording fixes

Esshahn:
another round of wording fixes, mostly lower case "du", "ihr", "deine" etc
We use lowercase writing for these from now on, e.g. "Teile dein Wissen"


vogelino (approved):
Top! Thx 🙏

dnsos (approved):

PR #113, opened 2021-09-13
feat(SensorCard): Add SensorCard, CategoryIcon and SensorSymbol comps

vogelino:
This PR adds 3 components:
- `SensorCard`, which represent a sensor and links to it
- `SensorSymbol`, which displays an SVG image based on a number between 1 and 32
- `CategoryIcon`, which renders an SVG icon based on a category ID


dnsos (approved):
Great implementation! I like how you accounted for different amounts of data in the sensor cards. Seems like a very robust UI.

Looking forward to seeing this included in the pages!

PR #116, opened 2021-09-14
Feat/new tabs component

vogelino:
This PR Adds a new Tabs component for later use in logged-in profile and tokens pages.


dnsos (approved):
Good work!

I noticed we still have the `RadioTabs` component around. We should not forget to delete old/deprecated components once they are completely removed from the UI - just a reminder for us.

PR #119, opened 2021-09-15
feat(SensorPageHeader): Add component for the header of the future sensor page

vogelino:
This PR adds a component for the header of the future sensor page.


public/images/sensor-page-header-pixels-mobile.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="336px" height="84px" viewBox="0 0 336 84" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>Artboard</title>
```
dnsos, line 3: There is some unnecessary stuff in the SVGs. You used Sketch, right? If so, there is a plugin called _SVGO Compressor_ that nicely removes everything that is unnecessary and gives a nice, clean SVG when exporting from Sketch.

dnsos (approved):
Looks amazing (on all screen sizes). Great work!

public/images/sensor-page-header-pixels-mobile.svg
vogelino: True! I normally use it but I forgot for this file. It will clean it! Thx! 🙏

PR #120, opened 2021-09-15
feat(DropdownMenu): Add ability to have disabled items in DropdownMenu

vogelino:
This PR adds the ability for DropdownMenus to include disabled items. This will be useful later when we will add the download link on the SensorPage.


dnsos (approved):
Good addition!

PR #121, opened 2021-09-16
fix(LandingBackgroundMap): Use Lat/Lng in home map instead of location string

vogelino:
This PR uses the lat/lng fields of the sensor in the landing page map instead of geocoding the location string.

It also changes the `TILESET` env variables to include one with labels and one without labels. 
@ff6347 @Esshahn @dnsos @julizet please update your `.env` file according to 1password entry. 🙏


src/components/LandingHeroBackgroundMap/index.tsx
```diff
@@ -75,7 +43,8 @@ export const LandingHeroBackgroundMap: FC<LandingHeroBackgroundMapPropType> = ({
         clickHandler={() => undefined}
         markers={[
           {
-            ...locationViewport,
+            latitude: sensor.latitude || 0,
```
dnsos, line 46: It feels wrong to have a fallback value for lat/lon.

@ff6347, perhaps we could already make this non-nullable in the DB?

dnsos (commented):
Great, thanks for the refactor @vogelino.

One question: We now have labels also on the landing page which was not the case before. I remember our consideration was that the labels might clash with the heading and text for this section. I actually think that the labels add nice context and maybe we can instead think of a solution where we move the section text somewhere else in the UI?

dnsos:
On second thought, the map labels now also clash with the hero section:
Screenshot 2021-09-16 at 11 20 29

Maybe we can remove the minimal transparency here?

vogelino:
> maybe we can instead think of a solution where we move the section text somewhere else in the UI?

I totally agree with that too. If I recall correctly, this is also one of the points that @julizet noted after testing the app with @Lisa-Stubert. I would also vote for moving the stats section somewhere else.

src/components/LandingHeroBackgroundMap/index.tsx
vogelino: I agree, I had to do it because the types indicated them as optional. I would vote for making them mandatory in the DB too.

vogelino:
> Maybe we can remove the minimal transparency here?

Oh yeah. I meant to do the opposite. No labels on top and some on the cards. I actually like the semi-transparency effect. Of course, we could also remove it completely.

dnsos:
> No labels on top and some on the cards.

Maybe that's the solution for now. At some point we could discuss how to add labels to the big map then.

vogelino:
> Maybe that's the solution for now. At some point, we could discuss how to add labels to the big map then.

I tried removing the transparency and to be honest, I think it works well like this. The transparency is not worth sacrificing the labels. Have a look when you can and let me know what you think @dnsos. Cheers.

dnsos (approved):
Absolutely, it doesn't lose anything and the labels are a really nice piece of information. Thanks for the update!

vogelino:
@dnsos I also added a background to the stats section so that we don't have collision with the background until we choose another solution. :)

src/components/LandingHeroBackgroundMap/index.tsx
ff6347: > perhaps we could already make this non-nullable in the DB?

Should we? What if the user does not know the location yet? What if he does not want to publish the location (maybe its in his home)? What if I want to prepare some sensors and position them later on? 

Also setting the lat/lon to 0/0 puts the sensor somewhere west of Afrika. So we shouldn't do that either. Keep them (null in the DB and undefined in the frontend and don't display them if lat or lon is missing.

dnsos: > Should we?

My impression so far was that a sensor without a geolocation is contrary to the idea of Stadtpuls. But let's not discuss this deep down in a PR review and instead talk about it next week in our checkin

PR #123, opened 2021-09-22
feat(STADTPULS-368): New Sensor Modal

vogelino:
This PR adds a Sensor modal component for the addition, edition, and deletion of a sensor.


src/components/EditAddSensorModal/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+        </fieldset>
+        <Controller
+          name='description'
+          control={control}
+          defaultValue={defaultValues?.description}
+          render={({ field }) => (
+            <FormTextarea
+              {...field}
+              label='Beschreibung'
+              placeholder='Beschreibe kurz der Sensor'
```
dnsos, line 188: Typo: der -> den

src/components/EditAddSensorModal/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+              defaultValue={defaultValues?.integration || integration}
+              render={({ field }) => (
+                <FormListBox
+                  {...field}
+                  onChange={newValue => {
+                    field.onChange(newValue);
+                    setIntegration(newValue as "http" | "ttn");
+                  }}
+                  label='Integration'
+                  placeholder='Wie möchtest du dein Sensor integrieren?'
```
dnsos, line 235: Typo: dein -> deinen

src/components/EditAddSensorModal/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+            <Controller
+              name='integration'
+              control={control}
+              defaultValue={defaultValues?.integration || integration}
+              render={({ field }) => (
+                <FormListBox
+                  {...field}
+                  onChange={newValue => {
+                    field.onChange(newValue);
+                    setIntegration(newValue as "http" | "ttn");
```
dnsos, line 232: Isn't it possible to set this to `as definitions["sensors"]["connection_type"]`? This way we don't have to manually update this once the `connection_type`s change

src/components/FormListBox/FormSelect.stories.tsx
```diff
@@ -0,0 +1,188 @@
+import { action } from "@storybook/addon-actions";
+import { Story, Meta } from "@storybook/react";
+import { useForm } from "react-hook-form";
+import { FormListBox, SelectOptionType } from ".";
+import ArrowBidirectionalVertical from "../../../public/images/icons/16px/arrowBidirectionalVertical.svg";
+import ArrowOutOfDoor from "../../../public/images/icons/16px/arrowOutOfDoor.svg";
+import Funnel from "../../../public/images/icons/16px/funnel.svg";
+
+export default {
+  title: "Forms/FormListBox",
```
dnsos, line 10: Great that this one is so flexible, to handle icons, default selections etc. 👍

src/components/FormListBox/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+      <div className={`${className} mb-2 relative`}>
+        <Listbox
+          value={options.find(({ value }) => value === selectValue)}
+          onChange={newVal => {
+            if (typeof onChange === "function") onChange(newVal?.value);
+            setSelectValue(newVal?.value);
+          }}
+        >
+          {label && (
+            <Listbox.Label className='block mb-2 cursor-pointer transition hover:opacity-60'>
```
dnsos, line 63: So this is how we use `headless-ui`? It gives us a completely unstyled element and we use Tailwind to style it?

src/components/SensorPageHeader/withData.tsx
```diff
@@ -0,0 +1,204 @@
...
+            "left-1/2 transform -translate-x-1/2 backdrop-filter backdrop-blur-md",
+          ].join(" ")}
+        >
+          {error && (
+            <Alert
+              type='error'
+              title='Fehler!'
+              message={
+                <>
+                  Es ist beim editieren ein Fehler aufgetreten:{" "}
```
dnsos, line 89: Maybe better: "Beim Editieren ist ein Fehler aufgetreten"

src/components/SensorPageHeader/withData.tsx
```diff
@@ -0,0 +1,204 @@
...
+                variant='secondary'
+                onClick={() => setDeletionConfirmationIsOpened(false)}
+              >
+                Abbrechen
+              </Button>
+            </>
+          }
+        >
+          <div className='prose max-w-none'>
+            <p>Du bist dabeit der Sensor „{mergedSensor.name}“ zu löschen.</p>
```
dnsos, line 194: Typos, should be: Du bist dabei, den Sensor [...] zu löschen.

dnsos (changes requested):
Great work @vogelino!  This really seemed like a complex one. I tested the interaction and it works very nicely.

I have a few comments and questions and one display problem:

Screenshot 2021-09-23 at 11 15 10

The tooltip turns slightly transparent and is outside of the viewport. The latter is of course a problem with the tooltip itself, not this PR.

src/components/EditAddSensorModal/index.tsx
vogelino: The problem is that the backend currently also accepts "other" but the frontend doesn't. I will omit "other".

vogelino: I tried omit but it isn't possible on enums... :(

src/components/FormListBox/index.tsx
vogelino: Exactly, and it takes care of all the aria labels, keyboard navigation, etc. Very cool!

src/components/SensorPageHeader/withData.tsx
vogelino: OMG all the typo... Sorry. Danke :D

vogelino:
@dnsos I fixed the opacity of the tooltip (not the position, I would do that in another branch) and added the focus on the name input on mount. That should do the trick. Let me know what you think?

dnsos (approved):
Good changes, thank you! 👍 

(I also think the tooltip positioning should be fixed in another task)

PR #124, opened 2021-09-23
fix(DropdownMenu): use purple text when item enabled

dnsos:
The dropdown items were only colored purple on hover, now they are by default.


vogelino (approved):
👍

PR #133, opened 2021-09-29
refactor(STADTPULS-393): New Registration UI (PART 2)

vogelino:
This PR is the continuation of #132 which was merged into main as a quick fix.

It includes a new signin route that uses the API's signin route to make sure that emails that aren't registered yet aren't allowed.


dnsos (approved):
👍

src/lib/requests/createSigningApiUrl/index.ts
```diff
@@ -0,0 +1,6 @@
+const API_VERSION = "v3";
+
+export const createSigningApiUrl = (resource: "signin" | "signup"): string =>
```
dnsos, line 3: Nice, that makes it a bit more obvious that the `NEXT_PUBLIC_TOKEN_API_URL` is in fact the same one. We could also think about renaming the Env Var to something more appropriate that includes all its functionalities.

Maybe consider fixing the mini typo here and wherever it's used: `createSigningApiUrl` -> `createSigninApiUrl`

src/lib/translationUtil/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+        Hast du noch keinen Account?{" "}
+        <TextLink href='/signup'>Registriere dich!</TextLink>
+      </>
+    );
+  if (errorMessage === 'body.email should match format "email"')
+    return "Die angegebene E-Mail-Adresse ist ungültig";
+  if (errorMessage === "Rate limit exceeded, retry in 10 seconds")
+    return "Du hast in einem kurzen Zeitraum zu oft versucht, dich einzuloggen. Bitte probiere es in 10 Sekunden noch einmal.";
+  if (errorMessage === "Rate limit exceeded, retry in 1 minute")
+    return "Du hast in einem kurzen Zeitraum zu oft versucht, dich einzuloggen. Bitte probiere es in eine Minute noch einmal.";
```
dnsos, line 17: Typo - > "in eine**r** Minute"

dnsos (approved):
Great work! The implementation is very clean and this will surely be a UX improvement. I especially like how you translated the error messages to something more understandable.

src/lib/requests/createSigningApiUrl/index.ts
vogelino: Thx! Yes it's true I thought about this. I'll make a suggestion. We have to remember updating the vars in vercel and 1password.

vogelino: I intentionally called the function `createSigningApiUrl` as it doesn't only produces urls for sign **in** but also for sign **up**. But maybe there is something less confusing?

src/lib/translationUtil/index.tsx
vogelino: Will change!

src/lib/requests/createSigningApiUrl/index.ts
dnsos: > I intentionally called the function createSigningApiUrl

Ahhh, got it... In that case, I am fine with leaving it the way it is. One alternative could be `createAuthApiUrl`?

vogelino: I now have removed the `createSigningApiUrl ` function entirely and refactored the different functions:
- `createSupabaseUrl` -> Former `createApiUrl`
- `createApiUrl` -> Now leads to our API

dnsos: Great, this is much more straightforward!

PR #134, opened 2021-09-30
refactor(STADTPULS-354): Cleanup old projects files

vogelino:
This PR: Removes everything related to the projects, including mock data, components, stories and tests.


dnsos (approved):
Great, thanks for the cleanup! Glad that all this old stuff is out of the way now.

PR #135, opened 2021-09-30
Various bugfixes + better magic link feedback

vogelino:
This PR:
- Fixes a text "Dein einzigartigen Nutzername" to "Dein Nutzername"
- Compares usernames for uniqueness regardless of their case to avoid problems in the URLs.
- Adds better feedback for magic links including small illustrations.


dnsos (approved):
Good updates, the case insensitiveness was quite important, indeed!

dnsos (approved):

PR #136, opened 2021-09-30
refactor(STADTPULS-327): exchange sensor API routes

dnsos:
This PR exchanges the API routes for the sensor records on the sensor page.

The routes from the frontend repo were removed and the backend ones used instead.

Also, the CopyToClipboard functionality was added to the API URL field.


vogelino (approved):
Great! Thx! 👍

PR #137, opened 2021-10-01
feat(Page Transitions): Add nextjs-progressbar lib for page transitions

vogelino:
This last PR before my vacation (very low effort, big benefit) adds a progress bar to the top of the page when page transitions occur. This should make the process of waiting for SSR calls to be finished and the page to be transitioned more understandable and enjoyable.

Love you all ❤️ 🌴 🏖️ See you soon!


Esshahn:
love it!

PR #139, opened 2021-10-05
feat(STADTPULS-430): redirect to user account after signin

dnsos:
This PR makes it possible to redirect to the user's account after clicking the magic link in the email:

- an `/account` route was added that first renders a "loading" profile
- we can set this route as the magic link redirect URL in Supabase
- once the user is authenticated, we redirect to the user's account `/account/[username]` (possible because we can retrieve the username from the `useAuth` hook)

Before merging:
- [ ] add new `/account` redirect URL to the production Supabase

### DIFF
--- a/pages/account/index.tsx
+++ b/pages/account/index.tsx
@@ -1,0 +1,55 @@
+import { FC, useEffect } from "react";
+import { useRouter } from "next/router";
+import { useUserData } from "@lib/hooks/useUserData";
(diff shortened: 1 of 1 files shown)


vogelino:
Very cool 😎! Well done! 

Cheers from Barcelona 😘

image

vogelino (approved):

PR #151, opened 2021-10-13
fix(STADTPULS-387): add spacing between menu items

dnsos:
WIP


dnsos:
Closing this as we decided to not fix. The `gap` property is supported in modern versions of both Safari and Chrome, so no point in supporting extremely old versions.

PR #155, opened 2021-10-28
refactor(Logout): Redirect after logout

vogelino:
This PR ensures that users are redirected after logging out.

PS: No tests were added because it happens in a tricky part and after many attempts to mock the right things I had to forfeit. I don't think it is that important for such a small change though.


ff6347:
> PS: No tests were added because it happens in a tricky part and after many attempts to mock the right things I had to forfeit. I don't think it is that important for such a small change though.

We should take a look into Testing with cypress, playwright et al for things like these. 

Mocking should be our last resort

dnsos (approved):
Yes, let's talk about other testing strategies for the future! I agree that mocking some stuff like this is cumbersome and feels wrong.

PR #156, opened 2021-10-28
refactor(integrationsUtil): Create integrationsUtil to gather integration constants

vogelino:
This PR extracts any reference to the integrations to an external file dedicated for this purpose. I didn't do a hook here as there is no particular advantage to using one here. I just use a file from which I export the constants.


dnsos (approved):
Straightforward refactoring that makes a lot of sense, thanks!

PR #157, opened 2021-10-28
refactor(UserInfoWithData): Apply random symbol id to new sensor modal

vogelino:
This PR adds a random sensor id to the new sensor modal form so that users don't necessarily have to think about this when creating a new sensor.

### DIFF
--- a/src/components/UserInfoHeader/withData.tsx
+++ b/src/components/UserInfoHeader/withData.tsx
@@ -16,6 +16,9 @@ interface UserInfoWithDataPropType {
   activeTab: "sensors" | "tokens";
 }
 
+export const getRandomSensorId = (): number =>
+  Math.round(Math.random() * (32 - 1) + 1);
+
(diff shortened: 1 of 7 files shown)


src/components/UserInfoHeader/withData.tsx
```diff
@@ -16,6 +16,9 @@ interface UserInfoWithDataPropType {
   activeTab: "sensors" | "tokens";
 }
 
+export const getRandomSensorId = (): number =>
+  Math.round(Math.random() * (32 - 1) + 1);
```
dnsos, line 20: Is there a way we can make this `32` less magic? Maybe assign it to a variable such as `NUMBER_OF_SYMBOL_SVGS`?

dnsos (approved):
Good addition, thanks!

dnsos (approved):
Amazing, thanks for the change!

PR #162, opened 2021-11-01
fix(DayPickerInput): Allow date input fields to enter date manually

vogelino:
This PR fixes the fact that date input fields wouldn't let dates be entered by typing, rather only by using the UI. This is now possible.


dnsos (approved):
Cool! The UI is very intuitive to use, but I see that some may prefer entering the date manually.

PR #163, opened 2021-11-03
fix(STADTPULS-438): max rows line chart

dnsos:
Before merge:
- [x] update Supabase max rows limit in production instance
---
This PR updates the line chart so that it can display more values. We found that **about 3000 values are the maximum** for our current charting library (visx). A `canvas`-based alternative should be explored in the future to increase the number of renderable values.

If the selected timespan has more values than the defined 3000, we show an alert informing about the limitation of the line chart.

Screenshot 2021-11-03 at 17 40 30

~~Note that it is still possible to select _Alle_ even for requests that exceed 3000 values. This is because we don't want to limit accessing more than 3000 values for the download button. The button should not be limited by the rendering limits of the line chart. The alert message is shown upon selecting an _Alle_ that exceeds 3000.~~

This PR also introduces a new environment variable `NEXT_PUBLIC_SUPABASE_MAX_ROWS` that can be used for ensuring we are always in sync with our settings in Supabase. Please refer to the `.env` values in the shared 1Password vault.


dnsos:
@vogelino, I will put this back to Draft because there was one thing I didn't consider. Will let you know once this is fixed.

dnsos:
> there was one thing I didn't consider

Nevermind, we talked about it and actually this is good to go. Reopening this!

vogelino (approved):
LGTM! Well done! Let's merge this! 👍

PR #168, opened 2021-11-09
Feat(STADTPULS-318): Make all data downloadable with Web Workers

vogelino:
This PRs defers downloads to a web worker to avoid blocking the main thread. It shows the downloads in the header. Each download can be canceled and clicking on a download leads to the relevant sensor page.


buildWorkers.js
```diff
@@ -0,0 +1,23 @@
...
+    onRebuild(error, result) {
+      if (error) console.error('watch build failed:', error)
+      else console.log('watch build succeeded:', result)
+    },
+  } : false,
+})
+  .then(() => { console.log("Build succeeded.") })
+  .catch((e) => {
+    console.log("Error building:", e.message);
+    process.exisst(1)
```
dnsos, line 22: Is `exisst` really a function? Or a typo?

src/components/CircularProgressBar/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+import colors from "../../style/colors";
+
+export interface CircularProgressBarPropType
+  extends Partial<CircularProgressbarProps> {
+  percentage?: number;
+  children?: ReactNode;
+}
+
+export const CircularProgressBar: FC<CircularProgressBarPropType> = ({
+  percentage = 66,
```
dnsos, line 17: Why default to 66?

src/lib/requests/getRecordsBySensorId/index.ts
```diff
@@ -1,6 +1,9 @@
 import { supabase } from "@auth/supabase";
 import { definitions } from "@common/types/supabase";
 
+const MAX_DOWNLOADABLE_RECORDS = 1000000;
+const NEXT_PUBLIC_SUPABASE_MAX_ROWS = 5;
```
dnsos, line 5: Shouldn't this be defined in an env variable? And set to something other than 5?

dnsos (changes requested):
Very nice work with the web workers. Also the UX feels great. Fantastic work overall!

I noticed, however, that a download of about 5000 records fails (without any error message, neither in the UI nor in the console). Test case was "All data" for [this sensor](https://stadtpuls-frontend-git-feat-stadtpu-e5a0ff-technologiestiftung1.vercel.app/sensors/2) - tested in Chrome and Firefox, macOS. The download first progresses in the UI, but when 100% are reached, no download is triggered. Very small subsets work, but it actually also happens when I try to download a subset of about 250 records. (Maybe it's somehow related to my browser settings?!)

One small visual flaw: When the download queue is in focus at the moment the progress reaches 100%, two visual artifacts remain on the screen, see screenshot:

Screenshot 2021-11-18 at 10 21 26

buildWorkers.js
vogelino: Typo! thx good catch!

src/components/CircularProgressBar/index.tsx
vogelino: True, I though I had changed that to 0. Will change! Thx

src/lib/requests/getRecordsBySensorId/index.ts
vogelino: True! I'm sorry, I think I'm quite tired... Will change!

vogelino:
> I noticed, however, that a download of about 5000 records fails (without any error message, neither in the UI nor in the console).

True! This happens in some cases where the progress reaches 100% but no CSV result is yet available. I added some check to make sure this doesn't happen anymore! Good catch!

> One small visual flaw: When the download queue is in focus at the moment the progress reaches 100%, two visual artifacts remain on the screen, see screenshot:

True! I added a check that makes sure that the caret is never displayed when the content of the dropdown is empty. Good catch as well!

dnsos (approved):
Very nice, thanks for all the changes! The downloads are working as expected now 💯

PR #171, opened 2021-11-15
refactor: build the homepage with ISR

dnsos:
This PR enables Next's [Incremental Static Regeneration (ISR)](https://nextjs.org/docs/basic-features/data-fetching#incremental-static-regeneration) for the homepage. This means:

- The homepage is generated at build-time instead of rendered by the server on every hit
- **Load times for the homepage are reduced**: In my tests the staging URL loads the DOM in 100 - 300 ms, while stadtpuls.com needs about 1 s
- The only "dynamic" elements on the homepage are regenerated every 60 s at most (see ISR docs for details). This is relevant mainly for the stats section. I simulated adding and deleting sensors and the ISR promise of regeneration works as expected.

Note that the effect of ISR [is not visible in development](https://nextjs.org/docs/basic-features/data-fetching#runs-on-every-request-in-development).

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -11,16 +11,16 @@ import {
   getLandingStats,
   LandingStatsReturnType,
 } from "@lib/requests/getLandingStats";
-import { GetServerSideProps } from "next";
+import { GetStaticProps } from "next";
 import { FC, useState } from "react";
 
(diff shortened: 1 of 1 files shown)


vogelino (approved):
Amazing! This is such an improvement! Thx a lot for this update!

PR #172, opened 2021-11-15
fix: make hero buttons clickable on small screens

dnsos:
This PR fixes the stacking of the hero section. On small screens (below Tailwind's `lg`) the hero image was positioned not next to but under the intro text. However, because the image has a `z-index` of 10, the buttons of the intro text were covered by the image and therefore not clickable. This PR lifts the intro text's `z-index` to 20 on small screens.


vogelino:
Great! Good catch! Thx for the good initiatives! 🎉

vogelino (approved):

PR #174, opened 2021-11-16
feat(STADTPULS-469): case insensitive account URLs

dnsos:
This PR makes the URLs for `/accounts/[username]` case-insensitive. ~~The solution is a bit hacky, but as far as I know Next doesn't have a better solution for making dynamic URLs case-insensitive (correct me if I'm wrong).~~

Todo:
- [x] don't simply add username from query to page props (wrong cases would be used)


pages/accounts/[username]/index.tsx
```diff
@@ -6,16 +6,24 @@ import { UserInfoWithData } from "@components/UserInfoHeader/withData";
...
 import { ParsedSensorType } from "@lib/hooks/usePublicSensors";
 import { useUserData } from "@lib/hooks/useUserData";
+import { getUserNames } from "@lib/requests/getUserNames";
 
 export const getServerSideProps: GetServerSideProps = async context => {
   try {
     const username = context.query.username;
     if (!username || Array.isArray(username)) return { notFound: true };
 
-    const accountData = await getAccountDataByUsername(username);
```
Comment (vogelino) [inline pages/accounts/[username]/index.tsx:15]: I don't think we need to get all usernames and compare them on the client-side. Instead, we should lowercase the username in the Supabase SDK call inside the `getAccountDataByUsername` function and use supabase's case insensitive syntax to find the relevant user `.ilike('column', '%CaseInsensitive%')` instead of `.eq('column', 'Equal to')`. This way, the comparison happens on the server. What do you think?

vogelino (changes requested):
Thx! However, I don't think this is the right approach, as we might have one million users in one year (yeaahhhhh, right...). See my comments below.

pages/accounts/[username]/index.tsx
Comment (vogelino) [reply pages/accounts/[username]/index.tsx:15]: See [here](https://app.supabase.io/project/anucvyvtfhtroefhotbr/api/default?resource=user_profiles) ans search for "case".

Comment (dnsos) [reply pages/accounts/[username]/index.tsx:15]: Good points, I've changed `getAccountDataByUsername` to use `.ilike("name", username)`. Note that I didn't add the `%`s as they would indicate that there could be more characters before or after the username. This way, we make sure that the comparison is strictly against the provided username - only case-insensitively.

---

A side note out of curiosity: To my understanding there wouldn't have been a client-side comparison of usernames with the previous solution since the call was inside `getServerSideProps`. It could have still led to long loading times with the surely expected 1 million users. But nothing client-side, or is there?

Comment (vogelino) [reply pages/accounts/[username]/index.tsx:15]: > But nothing client-side, or is there?

True! However, the initial load of the page would be slow because of that. Your newer version is much better.

vogelino (approved):
Perfect! Thx a lot! ❤️  Cleaner and more efficient! 👌

dnsos:
Thanks for the review @vogelino. I've made one more change that keeps the username from being passed to the page props with (potentially) wrong capitalization from the query. Feel free to take another look!

vogelino (approved):
👍

PR #179, opened 2021-11-18
new terms page

Esshahn:
++ adds a new page "Nutzungsbedingungen" to the documentation
++ adds new link to the new page in the footer
++ adds test for added footer link

Note: This is a first draft, likely to be extended later on. The position in the documentation is a suggestion, other ideas welcome.

### DIFF
--- a/pages/docs/terms.mdx
+++ b/pages/docs/terms.mdx
@@ -1,0 +1,21 @@
+---
+layout: 'docs'
+title: 'Nutzungsbedingungen'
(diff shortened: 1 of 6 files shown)


vogelino (commented):
Awesome! Thx! Even with a test! ;)
I noticed that the link in the footer isn't displayed right on mobile:

image

Maybe the footer should grow in height and elements should go to the next line. I can help you with that.

PR #183, opened 2021-11-23
feat(STADTPULS-501): less scary line chart alert

dnsos:
This PR updates the informational Alert component to have a light gray background and updates the content of the alert message when the line chart is asked to display more than 3000 records.


vogelino (approved):
Great change! Thx!

PR #184, opened 2021-11-23
chore(STADTPULS-500): use more appropriate filter label

dnsos:
Because we never display all records, only the maximum that is displayable.


vogelino (approved):
👍

PR #185, opened 2021-11-23
updated warning in docs

Esshahn:
++ added warning message in welcome page

### DIFF
--- a/pages/docs/index.mdx
+++ b/pages/docs/index.mdx
@@ -26,6 +26,8 @@ Drei Gründe, die für *Stadtpuls* sprechen:
 <br></br>
 <span className="w-12 h-12 rounded-full bg-gray-200" style={{padding: '0.3em 0.5em 0.3em 0.5em'}}>→</span> Die Nutzung der Plattform ist komplett <strong>kostenlos</strong>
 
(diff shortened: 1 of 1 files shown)


pages/docs/index.mdx
```diff
@@ -26,6 +26,8 @@ Drei Gründe, die für *Stadtpuls* sprechen:
 <br></br>
 <span className="w-12 h-12 rounded-full bg-gray-200" style={{padding: '0.3em 0.5em 0.3em 0.5em'}}>→</span> Die Nutzung der Plattform ist komplett <strong>kostenlos</strong>
 
+&#x26a0; Achtung: Stadtpuls befindet sich noch in einer frühen Entwicklungsphase, daher können noch Fehler auftreten und es an verschiedenen Stellen haken. Wir bitten euch daher um Verständnis und ein wenig Geduld. Wenn du Fragen, Feedback oder Anregungen hast, dann schreib uns eine E-Mail an: *info@citylab-berlin.org* oder nutze <a href="https://github.com/technologiestiftung/stadtpuls-community/discussions">Github Discussions</a>, um dich mit uns und der Community zu auszutauschen. Bugs kannst Du als <a href="https://github.com/technologiestiftung/stadtpuls-community/issues">Github Issue</a> reporten.
```
dnsos, line 29: Maybe instead of the Unicode symbol, we can use our custom warning icon to ensure a consistent display across clients?

It is located at `public/images/icons/16px/warningDuotone.svg` and can be used in an `` tag.

dnsos (changes requested):
Thanks, @Esshahn

I noticed that there is one capital "Du" which should probably be lowercased. Also, you're using "euch" in one sentence - should we maybe stick with "du" here?

pages/docs/index.mdx
Esshahn: good catch! I corrected the wording accordingly, thanks.
I kept the ascii code instead of the icon as styling this icon proved a bit too time consuming and I considered it not worth it right now (we use the code elsewhere, too)

dnsos (approved):

PR #186, opened 2021-11-23
feat(STADTPULS-488): sort and paginate accounts overview

dnsos:
This PR turned out a a bit differently than originally planned. STADTPULS-489 is outdated, instead we handle the accounts overview as discussed in the comments below.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,7 +9,7 @@
     "activityBarBadge.foreground": "#15202b",
     "statusBar.background": "#100C53",
     "statusBar.foreground": "#8330FF",
-    "statusBarItem.hoverBackground": "#8330FF",
+    "statusBarItem.hoverBackground": "#46ECA1",
     "titleBar.activeBackground": "#100C53",
     "titleBar.activeForeground": "#46ECA1",
     "titleBar.inactiveBackground": "#100C53",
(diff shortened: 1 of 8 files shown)


vogelino (commented):
Didn't we decide to put the accounts at the end of the list? Also, this solution is client-side. If we plan to add pagination this sorting won't work for long. We might rather do the sorting on the server-side. There is possibly a way to do this with supabase's `order` function right?

dnsos:
> Didn't we decide to put the accounts at the end of the list?
My understanding was that we would do that not for the accounts but for the sensors that don't have any records yet. But perhaps I'm wrong. @Esshahn, can you help out?

dnsos:
> If we plan to add pagination this sorting won't work for long.

Good point, thanks for the reminder. Maybe it makes sense to solve the pagination together with this

vogelino:
I think not showing accounts that exist is not so good. Making meaningful accounts easier to spot better. Rewarding instead of punishing. 🎖️

dnsos:
@vogelino or @Esshahn, please check again after the latest commits. I have now used this PR to actually implement the sorting and pagination of the accounts overview (same behaviour as the sensors overview).

vogelino (approved):
This is great! Thx for implementing!

PR #187, opened 2021-11-23
footer link feedback

Esshahn:
++ added feedback link in footer


vogelino (commented):
Good idea! Consider this approved but I will update the snapshots before merging.

PR #189, opened 2021-11-23
refactor(STADTPULS-475): Make default time filter be "max"

vogelino:
This PR sets the default time filter of the line chart on the sensor page as "All", because we know have better labels and a better indication for too data-intensive charts.


dnsos (approved):
Great, thanks!

PR #193, opened 2021-11-24
fix(STADTPULS-519): make accounts sorting case-insensitive

dnsos:
This is a quickfix to make the account sorting case-insensitive.

SQL's `ORDER BY` is  _case-sensitive_ by default, so is Supabase's implementation `sort()`. In an ideal world, we would add a custom SQL function, that we call via Supabase's `rpc()` method. That function would include a query like this:

```sql
SELECT * FROM user_profiles ORDER BY lower(name);
```

However, because of lack of time right now, I would suggest this quickfix that sorts accounts with JS, server-side for the accounts overview.


vogelino:
If I get it right, this only solves the problem within the results of a page itself? Also, I'm not sure how case sensitivity affects sorting? Why do we even need that?

dnsos:
Sorry, context: Yesterday I added [alphabetical sorting of the accounts](https://github.com/technologiestiftung/stadtpuls-frontend/blob/2d9b541395fba47ccca9ab180a8e93469ff7f8a2/src/lib/requests/getPublicAccounts/index.ts#L44) as discussed. However, this sorts by username case-sensitively (see attached screenshot). All usernames that start with caps are sorted first, all lowercase ones afterwards. But actually want alphabetical sorting that is case-insensitive, right?

Screenshot 2021-11-24 at 09 48 05

vogelino:
> But we actually want alphabetical sorting that is case-insensitive, right?

Right! But the solution that you provide only solves the problem for the currently displayed page. So you might still have some items in the wrong order appearing one page too early/late. This solution doesn't quite do the trick in my opinion. Shouldn't we wait to solve the problem with a solutions that doesn't require inaccurate frontend sorting? What do you think?

dnsos:
> Shouldn't we wait to solve the problem with a solutions that doesn't require inaccurate frontend sorting? What do you think?

Hm yes, I see your points and now also think that this is not the solution we need.

vogelino:
> Hm yes, I see your points and now also think that this is not the solution we need.

Hmm. Yeah. Sorry for the bummer... Maybe @ff6347 has an opinion on this. We might talk to the PostGresT API directly. Maybe there is a way to make sorting case insensitive there?

dnsos:
> Sorry for the bummer

Nah, thanks for the good review. I was moving a bit too fast with this one.

I would also like to hear @ff6347's opinion. I found one [blog post](https://dev.to/razvanstatescu/how-to-run-custom-sql-queries-using-functions-in-supabase-2nna) that explains using custom queries via `rpc()`

ff6347 (commented):
Didn't we implement that using a RPC https://github.com/technologiestiftung/stadtpuls-supabase/blob/b4b7a71ad5c5689dfb3badd728c3fe5ee921db98/supabase-docker-compose/dockerfiles/postgres/docker-entrypoint-initdb.d/50-remote-procedure-calls.sql#L20-L28

ff6347 (approved):
I guess we will replace this with a rpc in a future update

PR #194, opened 2021-11-24
fix(STADTPULS-517): Reduce humungous docs text size

vogelino:
This PR reduces the overall text size of the docs on larger screens.


Esshahn:
while it isn't huge anymore, I personally would prefer to have the text a bit smaller even.

vogelino:
> while it isn't huge anymore, I personally would prefer to have the text a bit smaller even.

I made it even smaller

dnsos (approved):
Nice, I like that size.

Esshahn:
cool

PR #195, opened 2021-11-24
fix(STADTPULS-521): fix Firefox glitch of disappearing grid

dnsos:
This PR removes the `float: left;` from the header of all pages except the `/docs`. While this seems to solve the display issue, I can't quite say why. It seems that using `float: left;` creates some positioning glitch (only in Firefox).

Do you have any idea why this happens @vogelino? Also, do you remember what was the reason for adding that style while building the beta banner?

### DIFF
--- a/pages/accounts/index.tsx
+++ b/pages/accounts/index.tsx
@@ -62,12 +62,12 @@ const AccountsOverview: FC<AccountsOverviewPropType> = ({
 
   if ((!accounts || accounts.length === 0) && pageIsWithinPageCount)
     return (
-      <div className='container mx-auto max-w-8xl py-24 px-4'>
+      <div className='container mx-auto max-w-8xl pt-12 pb-24 px-4'>
         <h1 className='flex justify-center mt-8'>Keine Accounts vorhanden</h1>
       </div>
     );
   return (
-    <div className='container mx-auto max-w-8xl py-24 px-4'>
+    <div className='container mx-auto max-w-8xl pt-12 pb-24 px-4'>
       <div
         className={classNames(
           "sm:mt-1 md:mt-2",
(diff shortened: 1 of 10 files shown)


vogelino:
> Do you have any idea why this happens @vogelino? 

No idea! Although I know that `float-left` is always a pain in the 🍑 

> Also, do you remember what was the reason for adding that style while building the beta banner?

Yes! This is because of the `BetaBanner` Component. We use float left to remove the space that CSS `sticky` creates. We might need to do this only on the home page and use `position: fixed;` on other pages as it was before... I can have a look at that. This one is tricky.

dnsos:
> Yes! This is because of the BetaBanner Component. We use float left to remove the space that CSS sticky creates. We might need to do this only on the home page and use position: fixed; on other pages as it was before... I can have a look at that. This one is tricky.

Interesting. I'm curious, let's talk about it!

dnsos:
@vogelino I applied the changes we discussed. Can you take a look?

vogelino:
It works well! Thx. There is, however, a problem on the sensor page where the map is supposed to go below the header. I guess we have to use `float-left` there too... 😕  See in the screenshot below

Screen Shot 2021-11-24 at 2 13 34 PM

dnsos:
> There is, however, a problem on the sensor page where the map is supposed to go below the header.

Phew yes, well spotted. I fixed that. Hopefully now it works for all pages

vogelino (dismissed):
Great! Everything works great now!

vogelino:
> The "How it works" section also disappears randomly in Firefox

Weird... This doesn't happen to me weirdly. Anyway, you might be right. I guess if we can find a solution without using `float-left` it would be way better. What about offsetting the hero block with a negative padding kinda like `calc(-1 * var(--headerHeight))`

dnsos:
Thanks for the final fix, @vogelino.

I re-requested another review, just for the sake of keeping a a proper process. However, if you have no time for that, I'm gonna merge this soon since we already agreed on the solution and it works.

vogelino (approved):
👍

PR #196, opened 2021-11-24
Better inline code styles in docs

vogelino:
This PR fixes the styles of the inline code elements in the docs.

**Before:**
Screen Shot 2021-11-24 at 4 52 32 PM

**After:**
Screen Shot 2021-11-24 at 4 51 40 PM


dnsos (approved):
Nice! This previous non-styling has always bothered me. Thanks for making it prettier!

PR #197, opened 2021-11-24
fix(STADTPULS-520): Remove spinner from loading bar

vogelino:
This PR removes the Loading spinner from the loading bar in the top as it always overlaps with other elements and there is no easy quick way to move it somewhere else. The progress bar should be enough to indicate progress.


dnsos (approved):
Agree, having only a loading bar seems like a quite common thing for many websites (including GitHub).

PR #205, opened 2021-11-29
chore(STADTPULS-468): truncate cards at 100 chars for description

dnsos:
This was missed in the first PR for STADTPULS-468.

### DIFF
--- a/src/components/AccountCard/index.tsx
+++ b/src/components/AccountCard/index.tsx
@@ -13,7 +13,7 @@ export interface AccountCardPropType {
   categories?: number[];
 }
 
-const DESCRIPTION_MAX_LENGTH = 50;
+export const DESCRIPTION_MAX_LENGTH = 100;
 
 const numberFormatter = new Intl.NumberFormat("de-DE", {
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
--- a/src/components/SensorCard/index.tsx
+++ b/src/components/SensorCard/index.tsx
@@ -14,7 +14,7 @@ export interface SensorCardPropType extends ParsedSensorType {
 }
 
 const DISPLAYABLE_RECORDS_AMOUNT = 20;
-const DESCRIPTION_MAX_LENGTH = 150;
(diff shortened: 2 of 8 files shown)


src/components/AccountCard/index.tsx
```diff
@@ -13,7 +13,7 @@ export interface AccountCardPropType {
   categories?: number[];
 }
 
-const DESCRIPTION_MAX_LENGTH = 50;
```
ff6347, line 16: Exporting the same thing twice in AccountCard/index.tsx and SensorCard/index.tsx
Would it make sense to have this in one place or are these lengths bound to visual space that is different on each card?

ff6347 (commented):
Looks. Good to me. See my question in the line comment

src/components/SensorCard/index.tsx
```diff
@@ -14,7 +14,7 @@ export interface SensorCardPropType extends ParsedSensorType {
 }
 
 const DISPLAYABLE_RECORDS_AMOUNT = 20;
-const DESCRIPTION_MAX_LENGTH = 150;
```
ff6347, line 17: Exporting the same thing twice in AccountCard/index.tsx and SensorCard/index.tsx
Would it make sense to have this in one place or are these lengths bound to visual space that is different on each card?

src/components/AccountCard/index.tsx
dnsos: Hmm, even though they have the same limit now, I would still not couple the values here.

The elements and the space on the cards are different, so it is possible that we need to separate them again when we touch the appearance of the cards. IMO, there is enough visual and semantic difference to the cards that justifies keeping the separation of the variables.

src/components/SensorCard/index.tsx
dnsos: See comment above ⬆️

ff6347: 👍🏽

ff6347 (approved):

PR #211, opened 2021-11-30
Fix env.example

ff6347:
env.example needs to point to localhost:3333 as well
related to #210 210


dnsos (approved):
Well spotted! Thanks for updating this.

PR #218, opened 2021-11-30
feat(STADTPULS-543): first story sensors on homepage

dnsos:
This PR brings the curated sensors to the homepage. We now display an intro to the first Stadtpuls Story and all related sensors on the map:

Screenshot 2021-12-01 at 17 47 30

Note that I've had to update the map view. It now doesn't span behind the other sections anymore but instead has a reduced height. This was necessary for being able to keep all markers in the map view, no matter the window size.

---

~~To Do:~~

- [x] double-check curated sensor IDs
- [x] prevent overlapping of markers and sensor cards
- [x] reduce height of map/story section in general
- [x] optional: add green sensor waves animation to markers
- [ ] ~~unrelated, but necessary: geolocation produced by sensor modal seems wrong~~ (-> different ticket)


Esshahn (approved):
Great work. I have created a new ticket with change suggestions though as I think we need to put more work into this. We should still merge this one though.

PR #220, opened 2021-12-02
updates docs

Esshahn:
merged and rewrote the welcome and about page so they are less redundant and streamlined.
also added new "mitmachen" page that we link to from the hero section


ff6347 (approved):
LGTM. I did not proof read it tough.

PR #221, opened 2021-12-02
Hero section text

Esshahn:
changes the hero section text. Not super happy with it, yet better than before.
Also changed button naming and the link to the docs.


ff6347 (changes requested):
On my 13'' Mac Book Pro the blocks from the image on the right overlap with the second button in the hero section and make it non clickable. Maybe the button needs a higher z-index ? 
Screenshot 2021-12-02 at 20 52 52

ff6347 (approved):

PR #229, opened 2021-12-07
refactor(STADTPULS-581): use DB view for accounts (fixes STADTPULS-496 + STADTPULS-519)

dnsos:
This PR heavily refactors the requests and hooks to use the newly available `extended_user_profiles` database view. This (hopefully) improves loading times, but most importantly fixes two bugs:

- the account cards are now sorted properly by alphabet, done on the DB side (STADTPULS-496)
- the account cards are not vulnerable anymore to high `recordsCount`s, the DB always provides the accurate count (STADTPULS-518)


src/lib/requests/getPublicAccounts/index.ts
```diff
@@ -1,24 +1,10 @@
...
-): number => {
-  if (a.username.toLowerCase() < b.username.toLowerCase()) {
-    return -1;
-  }
-  if (a.username.toLowerCase() > b.username.toLowerCase()) {
-    return 1;
-  }
-  return 0;
-};
+import { errors } from "../getPublicSensors";
```
ff6347, line 7: mixed import

Esshahn:
Looking good to me, tested:
- Account sorting
- Login
- Logout
- Edit Profile
- Add Sensor
- Add Token
- Remove Token

src/mocks/supabaseData/accounts.ts
```diff
@@ -1,30 +1,22 @@
...
-  AccountQueryResponseType,
-  mapPublicAccount,
-} from "@lib/hooks/usePublicAccounts";
-import { ttnSensors, httpSensors } from "./sensors";
-import { userprofiles } from "./userprofiles";
+import { definitions } from "@common/types/supabase";
+import { mapPublicAccount } from "@lib/hooks/usePublicAccounts";
 
-export const publicAccounts: AccountQueryResponseType[] = [
+export const rawUserProfiles: definitions["user_profiles"][] = [
```
ff6347, line 4: could be named just `userProfiles`?

src/mocks/supabaseData/accounts.ts
```diff
@@ -1,30 +1,22 @@
...
   {
     id: "1",
```
ff6347, line 6: id is a uuid/v4

ff6347 (changes requested):
Nice work. Looks good in the preview. All my comments are more stylistic then real errors. You decide which you want to address

src/lib/hooks/usePublicAccounts/usePublicAccounts.test.tsx
```diff
@@ -1,43 +1,23 @@
...
-    const onError = jest.fn();
-    const TestComponent = createTestComponent(onSuccessWrapper, onError);
-    render(
-      <SWRConfig value={{ dedupingInterval: 0 }}>
-        <TestComponent />
-      </SWRConfig>
+  it("returns the correct fields", () => {
+    const { result } = renderHook(() => usePublicAccounts());
+
+    expect(typeof result.current.accounts).toBe("object");
```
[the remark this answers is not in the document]
ff6347: `Array.isArray()` is better for arrays.

src/lib/requests/getPublicAccounts/index.ts
```diff
@@ -51,47 +37,25 @@ export const getPublicAccounts = async (
...
     if (error) throw error;
-    if (!data) return [];
-    const accounts = data
-      .map(mapPublicAccount)
-      .sort(alphabeticalAccountsSorter);
+    if (!extended_user_profiles) return [];
 
+    const accounts = extended_user_profiles.map(mapPublicAccount);
     return accounts;
   } else {
```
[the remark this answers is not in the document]
dnsos: Good idea! Check cde906b for the refactoring, @ff6347

src/lib/hooks/usePublicAccounts/usePublicAccounts.test.tsx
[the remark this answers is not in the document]
dnsos: Good point. I'm checking for `Array.isArray()` now (e149ee5)

src/lib/requests/getPublicAccounts/index.ts
```diff
@@ -32,30 +32,22 @@ export const getPublicAccounts = async (
...
-    if (error) throw error;
-    if (!extended_user_profiles) return [];
-
-    const accounts = extended_user_profiles.map(mapPublicAccount);
-    return accounts;
-  }
+  const defaultRangeStart = 0;
+  const defaultRangeEnd =
+    Number.parseInt(process.env.NEXT_PUBLIC_SUPABASE_MAX_ROWS as string) ||
+    1000;
```
ff6347, line 38: Aren't we on 3000 by default?

src/lib/hooks/usePublicAccounts/index.ts
```diff
@@ -39,34 +35,28 @@ export interface ParsedAccountType {
...
-  sensors,
   ...user
-}: AccountQueryResponseType): ParsedAccountType => ({
-  id: user.id,
-  username: user.name || "anonymous",
-  displayName: user.display_name || user.name || "Anonymous",
-  createdAt: user.created_at,
-  link: user.url,
+}: definitions["extended_user_profiles"]): ParsedAccountType => ({
+  id: user.id || "",
```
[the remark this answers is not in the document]
dnsos: All these fallbacks are necessary because the generated types are telling me that these fields may be `null`. I know that in reality this won't really be the case, but I didn't want to ignore the types.

Perhaps they are nullable because the DB view treats columns differently than "real" tables?

src/lib/requests/getPublicAccounts/index.ts
dnsos: 3000 is the amount of records the line chart can render. We are actually on `process.env.NEXT_PUBLIC_SUPABASE_MAX_ROWS`. The 1000 is just there as a fallback because _theoretically_ `NEXT_PUBLIC_SUPABASE_MAX_ROWS` can be undefined.

src/lib/requests/getAccountDataByUsername/getAccountDataByUsername.test.ts
```diff
@@ -1,24 +1,36 @@
...
       })
     );
     server.listen();
-    const accountData = await getAccountDataByUsername("bob");
+    const accountData = await getAccountDataByUsername(
+      exampleAccount.username as string
+    );
 
     expect(Array.isArray(accountData)).toBe(false);
     expect(accountData.id).toBe(exampleAccount.id);
```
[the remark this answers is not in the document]
dnsos: Yes, some of the tests are not ideal. I wouldn't improve them with this PR though, because (as far as I can see) they were all created earlier.

src/lib/hooks/usePublicAccounts/index.ts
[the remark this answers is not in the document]
ff6347: Might be the case. I take a look into pg views. Maybe we can fix this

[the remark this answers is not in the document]
ff6347: Hm. A view will not inherit the constrains of the source tables. ;(

src/lib/requests/getPublicAccounts/index.ts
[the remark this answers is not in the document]
ff6347: ❤️

src/lib/requests/getPublicAccounts/index.ts
ff6347: 👍🏽 
I normally create a module that does the checking of process.env. values for me
```ts
// env.ts
const NEXT_THING = process.env.NEXT_THING;
if(NEXT_THING === undefined) throw new Error('next thing is undefined')
export NEXT_THING
```

This is a little more work but it makes sure the values exist and I have a central place from where I import env vars. If I miss them I get an error.

ff6347: Also I don't scatter that many `parseInt (...) || 1` around the code

ff6347: But it is up to you.

pages/accounts/[username]/index.tsx
```diff
@@ -15,6 +14,7 @@ export const getServerSideProps: GetServerSideProps = async context => {
     return {
       props: {
         account: { ...accountData },
```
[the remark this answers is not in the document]
Comment (dnsos) [reply pages/accounts/[username]/index.tsx]: True. I actually did some refactoring here, now we've only got `sensors` inside of `account` (f17c7b7)

src/lib/requests/getPublicAccounts/index.ts
dnsos: > I normally create a module that does the checking of process.env. values for me

That's a really good idea. Though I wouldn't make this change in this PR because we have a rather time-critical moment.

But I'll definitely consider this for the future!

ff6347:
😻

dnsos:
@ff6347 I've addressed not all but most of your feedback - could you take another look?

src/lib/hooks/useUserData/index.ts
```diff
@@ -14,19 +14,15 @@ import {
   ParsedSensorType,
   SensorQueryResponseType,
 } from "../usePublicSensors";
```
ff6347, line 16: still a relative import here

ff6347 (changes requested):
LGTM. Just found one last relative import

src/lib/hooks/useUserData/index.ts
dnsos: True, thanks. Fixed now!

ff6347 (approved):

ff6347:
🚀  if all tests pass go ahead and merge!

PR #230, opened 2021-12-07
fix(STADTPULS-564): geolocation in sensor modal

dnsos:
This PR fixes the following issue:

When editing a sensor by dragging the map in the `EditAddSensorModal`, the updated coordinates were sometimes lagging behind the actual map viewport. To reproduce in production, you can try moving the map _slowly_ to position your sensor somewhere else. While doing this, it can occur that the viewport change is not updated in the fields that are submitted to the DB. This way the sensor location either remains the old one, or is pretty random on the map based on where the last viewport change was submitted to the fields.

**The fix**
`latitude` and `longitude` now have their own state and whenever one of them is (correctly) updated by `onViewportChange`, `setValue` is called to update the form fields. The value is now always correctly submitted.

### DIFF
--- a/src/components/EditAddSensorModal/index.tsx
+++ b/src/components/EditAddSensorModal/index.tsx
@@ -101,13 +101,10 @@ export const EditAddSensorModal: FC<EditAddSensorModalPropType> = ({
     viewport.latitude !== defaultValues?.latitude ||
     viewport.longitude !== defaultValues?.longitude;
 
-  const [latitude, setLatitude] = useState(viewport.latitude || DEFAULT_LAT);
-  const [longitude, setLongitude] = useState(viewport.longitude || DEFAULT_LNG);
-
   useEffect(() => {
-    setValue("latitude", latitude);
-    setValue("longitude", longitude);
-  }, [latitude, longitude, setValue]);
+    viewport.latitude && setValue("latitude", viewport.latitude);
+    viewport.longitude && setValue("longitude", viewport.longitude);
+  }, [viewport.latitude, viewport.longitude, setValue]);
 
   const {
     categories,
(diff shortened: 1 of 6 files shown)


Esshahn (changes requested):
lat/long update very smootly now, all looking good.
One little change request: on Safari, while I drag the map, the cursor "hand" symbol turns into a "edit text" symbol (vertical line).

vogelino (commented):
Great catch! I guess the Map component does a few optimizations, like throttling the `onViewportChange` method to avoid calling it too many times. I don't think, however, that we need additional `useStates` for `latitude` and `longitude`. Isn't it the `useEffect` that solves the problem? If this is true, the we should be able to keep using the `viewport` state and do something like this:

```js
useEffect(() => {
  setValue("latitude", viewport.latitude);
  setValue("longitude", viewport.longitude);
}, [viewport.latitude, viewport.longitude, setValue]);
```

Or did I get it wrong?

dnsos:
> I don't think, however, that we need additional useStates for latitude and longitude.

@vogelino true, that does the trick as well. I've changed the `useEffect` now to be more concise (6656d1f)

dnsos:
@Esshahn true. The cursor is actually set by Mapbox / React Map GL itself. See [one of their examples](http://visgl.github.io/react-map-gl/examples/controls) in Safari which has the same behaviour.

We can of course do some digging, but I would suggest to open a ticket for this to not block the fix for the geolocation. What do you think?

Esshahn:
agreed, makes sense

Esshahn (approved):

dnsos:
> but I would suggest to open a ticket for this

@Esshahn -> STADTPULS-611

PR #231, opened 2021-12-08
link fix for terms

Esshahn:
++ adds target blank to "Nutzungsbedingungen" link in signup modal.
++ adds register link to the signup modal

### DIFF
--- a/src/components/SignupForm/index.tsx
+++ b/src/components/SignupForm/index.tsx
@@ -173,7 +173,7 @@ export const SignupForm: FC<SignupFormPropType> = ({
               <>
                 Ich akzeptiere die{" "}
                 <TextLink
-                  target='blank'
+                  target='_blank'
                   href='https://www.technologiestiftung-berlin.de/de/datenschutz/'
                 >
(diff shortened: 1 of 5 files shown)


src/components/SignupForm/index.tsx
```diff
@@ -172,7 +172,10 @@ export const SignupForm: FC<SignupFormPropType> = ({
             label={
               <>
                 Ich akzeptiere die{" "}
-                <TextLink href='https://www.technologiestiftung-berlin.de/de/datenschutz/'>
+                <TextLink
+                  target='_blank'
```
dnsos, line 176: Just for completeness: Can you also add `rel='noopener noreferrer'`?

dnsos (changes requested):
Thanks @Esshahn! I have one tiny request (see below)

src/components/SignupForm/index.tsx
Esshahn: great catch, thx!

dnsos (approved):
Thanks!

PR #232, opened 2021-12-08
fix(STADTPULS-609): account tabs mobile

dnsos:
This PR improves the positioning of the tabs and the user stats on the user's account page on **very** small screens. See here for the improved tabs:

Screenshot 2021-12-08 at 11 31 54


Esshahn:
lacking such a device I just tried in the browser with mobile settings.

Esshahn (approved):

PR #233, opened 2021-12-08
fix(STADTPULS-601): Fix responsiveness to avoid annoying text-breaks

vogelino:
This PR changes the layout of the "How it works" section to avoid texts on mobile breaking badly and being unreadable.


dnsos (dismissed):
Great solution.

I noticed that there is no top spacing for this section:
Screenshot 2021-12-08 at 12 19 31

(But feel free to ignore, because it's not related to this ticket)

dnsos (approved):

PR #234, opened 2021-12-08
fix(STADTPULS-606): Adds CSSProperties

ff6347:
This adds a (IMO dirty) option to add additional CSSProperties to the ListBox.Options object
Adding the z-index via tailwind did not succeed.
Adding it to all instances makes it pointless.
Changing position to absolute breaks the layout.
Adding a z-index to higher up components did not succeed


dnsos (approved):
Maybe not the most satisfying fix, but it does the job and we have a reminder to fix this later on.

PR #236, opened 2021-12-08
refactor(STADTPULS-613): Story images on homepage slider

vogelino:
This PR adds images of the sensors's locations (like in story) instead of maps.

**WARNING:** The ids are those of production, therefore, this won't work on staging


dnsos (approved):
👍

PR #241, opened 2021-12-08
fix(STADTPULS-613): Make images staging-compatible

vogelino:
Makes image on home-slider staging compatible, meaning the ids of the fake sensors on staging can also be hard-coded in order to make the previews work.


dnsos (approved):
Nice, good improvement!

PR #242, opened 2021-12-08
Fix/records limit in account

dnsos:
This PR adds a limit to the records that are fetched when the user's sensors are fetched. In the refactoring yesterday, the limit got lost. Here it is reintroduced.

This fixes the problem that `getAccountDataByUsername` fetches records without limit and thus breaks the `accounts/[username]` route where the function is called and the user has too many records (e.g. 280.000).


ff6347:
Why do we fetch records at all when we want the sensors? If we cap it it makes no sense.

ff6347:
Agghhhhhhh. The chart on the card …

ff6347 (approved):
LGTM

vogelino (approved):
👍

PR #243, opened 2021-12-08
fix(STADTPULS-608): Make logo stop spinning at some point

vogelino:
This PR makes the logo animate only once so it can never endlessly spin when stuck on hover on mobile devices


dnsos (approved):
Great. The animation seems actually more elegant this way, also on non-mobile.

PR #245, opened 2021-12-09
fix(STADTPULS-526): Fix safari ovelapping elements in story block

vogelino:
Fix safari overlapping elements in story block

Screen Shot 2021-12-09 at 1 27 50 PM


dnsos (approved):
Great, thanks for the fix @vogelino

PR #248, opened 2021-12-09
fix(STADTPULS-535): Trim name of sensor, device and user everywhere

vogelino:
To avoid problems when communicating with the db, we trim the user-, ttnDevice- and sensorname everywhere.

### DIFF
--- a/src/components/UserInfoHeader/index.tsx
+++ b/src/components/UserInfoHeader/index.tsx
@@ -43,7 +43,7 @@ export const UserInfoHeader: FC<UserInfoHeaderPropType> = ({
         <div className='absolute top-0 left-0'>
           <span className='inline-block shadow rounded-full'>
             <UserAvatar
-              username={username}
+              username={username.trim()}
               size={windowWidth && windowWidth > 640 ? 80 : 48}
             />
           </span>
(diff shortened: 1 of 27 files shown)


src/components/UserInfoHeader/index.tsx
```diff
@@ -63,16 +63,16 @@ export const UserInfoHeader: FC<UserInfoHeaderPropType> = ({
...
-              href={`/accounts/${username}`}
+              href={`/accounts/${username.trim()}`}
               className={[
                 "text-purple hover:text-blue transition-colors",
                 "inline-block order-2 sm:order-3",
                 "mt-0 sm:mt-3",
               ].join(" ")}
             >
-              @{username}
+              @{username.trim()}
```
dnsos, line 75: I'm wondering if we need the trimming in places like this? If we ensure that it is not possible anymore to submit a username/sensor name etc. with whitespaces, shouldn't this be unnecessary?

dnsos (commented):
Thanks @vogelino. 

I'm wondering whether we need the trimming in places where we only "display" or use one of the mentioned fields, such as for the display of the user name? Doesn't it suffice to make sure the trimming happens in the requests to the DB?

src/components/UserInfoHeader/index.tsx
vogelino: This is just to avoid displaying spaces in case there are spaces present in the value. This is just a safety measure. While it should happen, it doesn't hurt. Do you think I should rather not do that?

dnsos: Thanks for the in-person the discussion and the new changes. Happy that the trimming now happens only where we connect to the database 👍

dnsos (approved):
Great, thanks!

PR #249, opened 2021-12-13
refactor(STADTPULS-625): Use definitions package

vogelino:


src/lib/hooks/useUniqueUsernameValidation/index.ts
```diff
@@ -7,7 +7,7 @@ const checkIfUsernameIsUnique = async (username?: string): Promise<boolean> => {
   const { data: usersWithMatchingUsernames, error } = await supabase
     .from<definitions["user_profiles"]>("user_profiles")
     .select("id")
-    .ilike("name", `%${username}%`);
+    .ilike("name", `%${username.trim()}%`);
```
dnsos, line 10: Why is the `trim` introduced here?

dnsos (commented):
Cool, it's nice that the types are now centralized!

There are a few `trim`s introduced. Why was there a need to have them with the new definitions?

src/lib/hooks/useUniqueUsernameValidation/index.ts
vogelino: This makes it defensive against leading and trailing spaces in the usernames. In a perfect world, this should not be necessary but with this additional step, no problem of leading/trailing slashes should occur.

dnsos (approved):

PR #250, opened 2021-12-14
fix(STADTPULS-628): Break and linkify long strings in sensor cards and page header

vogelino:
This PR does the following:
- On the sensor cards:
  - Break words that are too long
  - Show links as underlined spans (not clickable)
- On the sensor page headers:
  - Break words that are too long
  - Show links as blue, clickable links
  
  To test see Sensor with id: 61 (Should also be the first in the sensors overview page on staging)

Screen Shot 2021-12-14 at 11 29 45 AM
Screen Shot 2021-12-14 at 11 29 50 AM


dnsos (approved):
Good improvement!

PR #255, opened 2022-01-06
feat(STADTPULS-639): Implement list display of sensors

vogelino:
This PR implements the denser, sensors list as follow:

List view


ff6347:
Very nice. That was pretty quick. 🚀 

Some things I noticed:

- user names should be left aligned to their column
- What happens with very long usernames?
- in the screenshot on (done in staging env) the first row has a different offset then the following due to long descriptions and what not
- The animation for the hover highlight could be quicker. So only one is highlighted at the time
- The small chart has a small line at the end
- should there be a chart when there are no values?
- The value above the chart is the last value right? On first glance I thought it was the number of records.
- Icon and username a center aligned / icon and category are top aligned

Esshahn:
Not sure if intended: I think the sensor names (leftmost) are too big. I would reduce their size quite a bit to the size of the rest of the text in the row. It could still be bold
Bildschirmfoto 2022-01-06 um 13 22 29
.

vogelino:
Thx @ff6347 and @Esshahn a lot for your review! Good points!

---

> user names should be left aligned to their column

AND

> in the screenshot on (done in staging env) the first row has a different offset then the following due to long descriptions and what not

This only happens in extreme cases in order to maximize the number of displayed characters (Here it's the long aaaaaa word). In 99% of the cases, this doesn't happen.

---

> What happens with very long usernames?

You can see this case using the storybook story. It looks like this:


https://user-images.githubusercontent.com/2759340/148385543-d0fcabc2-0697-46f3-a174-70715751ed9a.mov

---

> The animation for the hover highlight could be quicker. So only one is highlighted at the time

I'm not sure what you mean. I don't see multiple things highlighted at once when hovering. Could you please clarify? Thx

---

> The small chart has a small line at the end

This is intended. This shows that the value displayed in the top right corner relates to the last value in the chart.

---

> should there be a chart when there are no values?

Hmm.. I think both is ok. It visually shows that there isn't any data. There also is the text "Keine Daten". I'm ok with both.

---

> The value above the chart is the last value right? On first glance I thought it was the number of records.

The gray line is supposed to help with that. But maybe this isn't enough. Any suggestion on how to solve this?

> Icon and username a center aligned / icon and category are top aligned

True! Good catch!

> Not sure if intended: I think the sensor names (leftmost) are too big. I would reduce their size quite a bit to the size of the rest of the text in the row. It could still be bold

I have reduced the font size

ff6347:
> I'm not sure what you mean. I don't see multiple things highlighted at once when hovering. Could you please clarify? Thx

I mean the reaction time of the hover effect is very slow if I move over the rows. See the gif

ff6347:
> This only happens in extreme cases in order to maximize the number of displayed characters (Here it's the long aaaaaa word). In 99% of the cases, this doesn't happen.

Can you try it with real german text. Since german tends to have way longer words then english

vogelino:
> I mean the reaction time of the hover effect is very slow if I move over the rows

I completely removed the transition. It was configured to last only 150ms, which rather indicate some inefficiency on the performance side of things. I think without animation it looks just snappy and sharp.

> Can you try it with real german text. Since german tends to have way longer words than English

I did. You can test the behavior yourself in the storybook. Here is a video:

https://user-images.githubusercontent.com/2759340/148405192-99791a4d-d7fc-44cc-975d-afb24287f0d1.mov

dnsos (approved):
This is really nice! The look of the new page is much better than the previous one and I like your implementation as well. 

Also, the previous discussions resulted in good improvements - I have nothing to add.

vogelino:
> This is really nice! The look of the new page is much better than the previous one and I like your implementation as well.
> 
> Also, the previous discussions resulted in good improvements - I have nothing to add.

Thx! Glad you like it! :)

PR #256, opened 2022-01-06
feat(STADTPULS-640): Change route of account and sensor to /[username] and /[username]/[id] resp.

vogelino:
- **/accounts/[username]** becomes **/[username]**
- **/sensors/[id]** becomes **/[username]/[id]**

The redirects are taken care of using NextJS' redirect feature. (These are not reflected in the sitemap.xml, only new routes will be present there)

Certain words that can conflict with current, previous, or future routes are now not allowed as usernames while creating an account to prevent unreachable routes.

Note that having incrementing ids on a per-user basis is **not possible** if we want to support redirects from old URLs as we cannot determine whom an ID would belong to if we stumble an URL such as `/sensors/12`. The way we do it now is by fetching the user by ID and then comparing adding its username to the URL.


ff6347:
This is awesome. Makes way better URLs. Bravo 👏🏽 

> Note that having incrementing ids on a per-user basis is not possible if we want to support redirects from old URLs as we cannot determine whom an ID would belong to if we stumble an URL such as /sensors/12. The way we do it now is by fetching the user by ID and then comparing adding its username to the URL.

I know an programmatic approach is preferable, but couldn't we just have a look up table (could also be a JSON that we maintain) for the sensor redirect? Since we don't have that many redirects

vogelino:
Thx! Appreciated! We could yes. I'm not sure if the effort is worth it. It is yet another call to make before returning the page. Also, do you know how we would implement the ids being incremental on a per user basis? Is there an SQL feature for this?

Esshahn:
thank you so much for implementing this <3 !

vogelino:
@ff6347 and I misunderstood each other. Lookup tables are not needed as IDs are already unique and can be used to construct a new URL that includes the username (by first fetching the sensor, grabbing its author username, and then returning a 301 redirect). Also, I understood that we wanted to have incremental ids for each user so that when john doe creates its third sensor, it becomes ID 3 and not whatever the last ID is in the whole table is. We agreed that this is unnecessary and overkill. Therefore, this PR is ready to be merged into staging.

ff6347 (commented):
LGTM. I'will have to test some things when it lands in staging, since I can't log in to the dev preview right now

ff6347 (approved):

ff6347:
Sorry. Wrong button. Approved

PR #257, opened 2022-01-10
refactor(STADTPULS-645): Restructure URL to include "sensors" in route

vogelino:
This PR adds `/sensors` to user-specific routes. Now the routes are as follows:
- `/username`: Redirect to `/username/sensors`
- `/username/sensors`: The user's sensors
- `/username/tokens`: The user's tokens
- `/username/ID`: redirect to `/username/sensors/ID`
- `/username/sensors/ID`: A specific sensor


ff6347:
NIce. 28 files change for a single word in the URL. What an age we live in.

Do we need this redirect? Since that is not in production yet?

```
/username/ID: redirect to /username/sensors/ID
```

pages/[username]/index.tsx
```diff
@@ -1,55 +1,20 @@
...
-        ) : (
-          <p>Keine Sensoren vorhanden</p>
-        )}
-      </div>
-    </>
-  );
+  return {
+    redirect: {
+      destination: `/${username}/sensors`,
+      permanent: true,
```
Comment (dnsos) [inline pages/[username]/index.tsx:12]: Does the permanent redirect make sense here? Based on @Esshahn's comment earlier today `username/index` might actually turn into a real page one day.

vogelino:
> NIce. 28 files change for a single word in the URL. What an age we live in.

Hahaha. Yeah well, those are mostly links `href` that lead to the new page. Nothing too crazy, fortunately!
 
> Do we need this redirect? Since that is not in production yet?

Not necessarily but I find it more helpful than sending a 404. Don't you think?

pages/[username]/index.tsx
Comment (vogelino) [reply pages/[username]/index.tsx:12]: But it will be completely new content. Therefore it isn't a temporary redirect either. It isn't a previously available page that will be reinserted later on. I wouldn't bet my life on this though. I'm really not sure.

Comment (vogelino) [reply pages/[username]/index.tsx:12]: I did some research and it seems you are right! This particular page will potentially be used in the future and we should therefore not do a permanent redirect. I will change that to be a temporary redirect.

ff6347:
> Not necessarily but I find it more helpful than sending a 404. Don't you think?

No. Since the ID is a generic thing and can mean anything it is better to give a 404.

vogelino:
> > Not necessarily but I find it more helpful than sending a 404. Don't you think?
> 
> No. Since the ID is a generic thing and can mean anything it is better to give a 404.

I'm not sure to understand. I get that in theory, the route is generic but when the URL matches a sensor, why not redirect the user to this sensor. It provides a much better experience for the user. Imagine someone enters the ID in the URL and forgets the `/sensors` prefix. This person would be redirected to the right place. If this doesn't match a real ID or an ID that belongs to the user in the URL, then of course a 404 would be returned. But directly having a 404 would not help the user at all to find their way. I don't see why this is better.

vogelino:
I removed the redirects from `/username/ID` to `/username/sensors/ID`, as deemed unnecessary by @ff6347. I'm still not convinced that this is better from a UX perspective. However, unless @dnsos or @Esshahn expresses a strong opinion in favor of having a redirect, I don't mind removing it.

dnsos (approved):
I'm split once again.

I see that generally it could be a good UX to have the redirects. However, I don't see the basis for assuming they are necessary. I'm thinking that if someone is able to enter a correct `/[username]/` into the URL, they would know of the `/sensors/` part or even have their browser suggest that path.

No strong pro-redirects from my side, in that sense I'm approving this PR!

PR #259, opened 2022-01-13
feat(STADTPULS-632): Sensors map

vogelino:
This PR implements the new split Sensors map, as designed in [STADTPULS-155](https://citylab.atlassian.net/browse/STADTPULS-155):

image

**WIP:** This is a work in progress. Please ignore for now.

**PS:** As part of the refactoring, the changes made include a bonus. The map on the landing page can now be clicked and this changes the active element in the slider.


vogelino:
I added e2e Tests for the sensors map. These work locally but no within the GitHub action. As discussed with @dnsos, there are 2 possible reasons we can think of now:
- [x] The GH action requires access to the `NEXT_PUBLIC_MAPBOX_TOKEN` env var (Added via secrets)
- [ ] The DB used for the GH action uses a locally spawned supabase DB which is empy and needs to be seeded (@dnsos is working on that in #266 

Until we fix that we won't be able to merge this. I'll have a look with @dnsos today.

src/components/SensorsList/index.tsx
```diff
@@ -8,7 +8,7 @@ interface SensorsListType {
 
 export const SensorsList: FC<SensorsListType> = ({ sensors }) => {
   return (
-    <ul className='flex flex-col'>
+    <ul className='flex flex-col w-[calc(100%+16px)] ml-[-8px]'>
```
dnsos, line 11: Just out of curiosity: what are these new width and margin values for?

src/components/SensorsMap/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+    document.documentElement.scrollHeight -
+    window.outerHeight -
+    footerHeight +
+    headerHeight;
+  const offsetPosition =
+    elementPosition + window.pageYOffset - headerHeight - paddingTop;
+
+  window.scrollTo({
+    top: Math.min(offsetPosition, maxScrollDist),
+    behavior: "smooth",
```
dnsos, line 60: How about only adding `"smooth"` if the `prefers-reduced-motion` media feature is not set to `reduce`? This way we would honour a user's preferences regarding motion sickness.

This would also apply to the `flyTo` of the map then...

tailwind.config.js
```diff
@@ -55,7 +56,7 @@ module.exports = {
       },
       animation: {
         borderpulse: "border-pulse 3s ease-in-out infinite",
-        "borderpulse-blue": "border-pulse-blue 3s ease-in-out infinite",
+        "bordezrpulse-blue": "border-pulse-blue 3s ease-in-out infinite",
```
dnsos, line 59: The "z" slipped in by mistake?

dnsos (changes requested):
Thanks for this great improvement, @vogelino! The page feels great on both mobile and desktop. Also the implementation looks very good.

I have a few comments and questions. Of those I would especially want to discuss the idea to honour the `prefers-reduced-motion` feature, if set by a user. Some of the animations and `flyTo`s are pretty wild, especially when jumping from intercontinentally.

One other thing I noticed is that the geolocation button is not working for me, neither on macOS 12.1, nor on Android 12.

cypress/integration/sensors.test.ts
```diff
@@ -1,7 +1,158 @@
...
+
+    cy.findByRole("textbox").type("Platz d. Luftbrücke 4, 12101 Berlin, Germany");
+
+    cy.get("[data-cy='map-search-result']")
+      .first()
+      .click({ waitForAnimations: false });
+    
+    cy.get("[data-cy^='marker-circle']").should("have.length", 3);
+  });
+  it.only("should show a sensor thumbnail when marker is clicked on mobile", () => {
```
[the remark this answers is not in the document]
vogelino: Yes! This only happens on mobile! Regarding the test with the geo search field, there isn't any particular reason. There isn't neither any reason to run it on desktop only, and duplicating it feels wrong. Do you suggest we move it to desktop?

src/components/SensorsList/index.tsx
vogelino: `w-[]` is a way from tailwindcss JIT to render arbitrary values. `calc(100%+16px)` adds 16px to a percentual 100% width. This makes it unnecessary to use the style attribute and makes leveraging the media queries and other modifiers such as `hover` possible.

src/components/SensorsMap/index.tsx
vogelino: Great idea! I will try to implement it!

tailwind.config.js
vogelino: It seems so! ':D I will fix it!

vogelino:
> One other thing I noticed is that the geolocation button is not working for me, neither on macOS 12.1, nor on Android 12.

Right! It seems to be the case for me as well! I will try to fix it! Thx! 🙏

cypress/integration/sensors.test.ts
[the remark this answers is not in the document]
dnsos: Ah no, maybe it's a misunderstanding. I just noticed the `it.only` here which causes the other two tests to not run at all:

- "should allow to switch between list and map"
- "should be able to search for a place"

I was wondering if it is on purpose because e.g. they are currently not working?

tailwind.config.js
vogelino: I removed the borderpulse-blue and bg-pulse animation as they weren't used anymore.

src/components/SensorsMap/index.tsx
vogelino: I did add a react hook called `useReducedMotion` and set the flyTo and scroll behavior accordingly

vogelino:
> > One other thing I noticed is that the geolocation button is not working for me, neither on macOS 12.1, nor on Android 12.
> 
> Right! It seems to be the case for me as well! I will try to fix it! Thx! 🙏

I did misinterpret the compass functionality for the geolocation button and forgot to test it. I now changed it and used the correct GeolocateControl component from react-map-gl.

cypress/integration/sensors.test.ts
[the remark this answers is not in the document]
vogelino: Oh! Yes! You're right. I removed the only keyword and now all tests are running

dnsos (approved):
Good changes and fixes! Thanks for implementing the `prefers-reduced-motion` feature, amazing!

PR #261, opened 2022-01-18
test(STADTPULS-622): add basic Cypress testing

dnsos:
This PR adds [Cypress](https://docs.cypress.io) for integration testing. This includes the configuration of TypeScript and ESLint to work with Cypress. 

Some basic test stubs were added as examples. 

---

[This article](https://www.valentinog.com/blog/canvas/) may be a good resource to start with testing `canvas` elements for our maps. EDIT: The article suggests using `cypress-plugin-snapshots` which [may be an unmaintained project](https://github.com/meinaart/cypress-plugin-snapshots/issues/194).


ff6347:
Awesome. Love it. Is it mandatory that the cypress files are called _spec? I would vote for .test as well. What do you think?

dnsos:
> Is it mandatory that the cypress files are called _spec? I would vote for .test as well

Thanks for the feedback and good idea, @ff6347. `_spec` is the Cypress default but it's absolutely possible to use something else. I renamed the files in 55e3480.

ff6347:
You need to make sure jest ignores the cypress folder

dnsos:
> You need to make sure jest ignores the cypress folder

Yep, Jest already doesn't care about it

vogelino:
Great addition! This should make testing interfaces and user flows much easier. Thx for taking care of that! 👍

vogelino (approved):

ff6347 (changes requested):
Awesome great addition. Looking forward to see this grow. Would be nice to have the cypress tests as on step in the main build.yml https://github.com/technologiestiftung/stadtpuls-frontend/blob/main/.github/workflows/build.yml and not as a separate file.

dnsos:
@ff6347, I added the Cypress tests to a `main.yml` workflow. I also did some refactoring to shorten the duration of this workflow by only using one build step instead of two. Let me know if you have advice for shortening the workflow even more.

I also suspect that soon we'll have integration tests that need a lot of time, specifically the ones that access pages with long loading times such as `/sensors` or `/accounts`.

.github/workflows/main.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, 'skip ci')"
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [14.18.0]
```
ff6347, line 16: Mayne this should be automated?

https://github.com/actions/setup-node/issues/32#issuecomment-525791142

```yml
- name: Read .nvmrc
  run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
  id: nvm

- name: Use Node.js (.nvmrc)
  uses: actions/setup-node@v1
  with:
    node-version: "${{ steps.nvm.outputs.NVMRC }}"
```

cypress/integration/home.test.ts
```diff
@@ -0,0 +1,7 @@
+describe("The Home Page", () => {
+  it("successfully loads", () => {
+    cy.visit("/");
+
+    cy.findByRole("link", { name: /Sensoren ansehen/i }).should("exist");
```
ff6347, line 5: If possible we should use findByText in these cases. Since this is what the user actually sees. 
https://github.com/testing-library/cypress-testing-library/blob/310c394ed7c678fe4e6f6f582d258de62bf6a669/types/index.d.ts#L66

.github/workflows/main.yml
ff6347: 🚀

ff6347 (approved):
Great addition @dnsos

cypress/integration/home.test.ts
ff6347: Since these are just stub tests it's fine for now.

dnsos: I actually use `*ByRole` intentionally since this way we test against the accessibility tree and not just for users who access Stadtpuls visually. I'm aware that we are far from being accessibility-friendly, but it's good to approach this direction. Btw, this is the [recommended query priority by Testing Library itself](https://testing-library.com/docs/queries/about#priority).

ff6347:
> I actually use `*ByRole` intentionally since this way we test against the accessibility tree and not just for users who access Stadtpuls visually. I'm aware that we are far from being accessibility-friendly, but it's good to approach this direction. Btw, this is the [recommended query priority by Testing Library itself](https://testing-library.com/docs/queries/about#priority).


Ah. Okay. Cool. Didn’t know that.

vogelino (commented):
Great additions indeed! Well done!

PR #268, opened 2022-02-14
fix(Next.js-config): Add mapbox tiles URL to content security policy

vogelino:
This PR adds the tiles.mapbox.com URL to the content security policy rules. This URL seems to be newly used by mapbox to host tiles and wasn't formerly allows in our content security policy. This might be a problem in the future.


dnsos (approved):
👍

PR #270, opened 2022-02-17
deps(Next.js): Upgrade next.js to version 12

vogelino:
This PR updates NextJS to the version 12. It seems that wenn using `swcMinify: true` some magic tree shaking or build process breaks mapbox on the deployed version. This might be fixed later on, or might require specific changes in the configuration. Not sure here. For now sticking to terser.


dnsos (approved):
Good work, thanks for upgrading @vogelino.

While you were fixing the CI, I tested the site from a user's perspective. There is no difference noticeable, which is very good. I admit that I couldn't look in **detail** into all the changed files. Perhaps we can talk later what were the important changes that had to be done.

vogelino:
> Good work, thanks for upgrading @vogelino.
> 
> While you were fixing the CI, I tested the site from a user's perspective. There is no difference noticeable, which is very good. I admit that I couldn't look in **detail** into all the changed files. Perhaps we can talk later what were the important changes that had to be done.

Thx! No changes should be noticeable for the users (excepted maybe speed differences due to the change of compiler).

PR #273, opened 2022-02-22
fix(MarkerMap): Assign different ids to markers and to clusters

vogelino:
This PR ensures that clusters and single markers never have the same IDs. This issue provoked some markers not being updated as they shared duplicated IDs.

PS: I also had to fix the function `getRecordsBySensorId` which failed in some tests when given no option (a case that was not used anywhere in business logic, so I deleted this branching and set an empty as default options) .


dnsos (approved):
Good finding and fix! 👍

PR #281, opened 2022-03-23
fix(STADTPULS-732): Fix requests failing because of sorting of inexistant records

vogelino:
This PR fixes the following type of errors, which are responsible for the incapacity of users to create sensor or to see the sensors in the home page:
```json
{
  message: "Cannot apply filter because 'records' is not an embedded resource in this request",
  hint: "Verify that 'records' is included in the 'select' query parameter."
}
```

The problem was that we are not querying the records in the sensors queries, but are still including sorting of records in the query. At the time this was done, no error was triggered. It seems supabase changed this and now throw errors. 

**PS:** The GitHub Action "Prevent merge from non-staging to main" will fail, as I am comparing this branch directly to main. In this case this is intended as it is a quickfix. Ignore the failing check.

### DIFF
--- a/src/components/SensorPageHeader/__snapshots__/SensorPageHeader.stories.storyshot
+++ b/src/components/SensorPageHeader/__snapshots__/SensorPageHeader.stories.storyshot
@@ -288,7 +288,7 @@ exports[`Storyshots Layout/SensorPageHeader Maximum Props 1`] = `
           onClick={[Function]}
           readOnly={true}
           type="text"
-          value="api.stadtpuls.com/api/v3/sensors/1/records"
+          value="localhost/api/v3/sensors/1/records"
         />
         <div
           aria-hidden={true}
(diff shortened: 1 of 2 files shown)


ff6347 (changes requested):
PR should go against staging and the tests are broken and github is having issues as well

vogelino:
> PR should go against staging and the tests are broken and github is having issues as well

As stated in the description of the PR 👆, as this is a quickfix, it compares to main and not to staging. Otherwise other changes made to staging would also be merged into main. The only tests failing are the snapshots, which I generated using the wrong .env file. I just fixed it and am waiting for the action to run once again.

ff6347:
> > PR should go against staging and the tests are broken and github is having issues as well
> 
> As stated in the description of the PR 👆, as this is a quickfix, it compares to main and not to staging. Otherwise other changes made to staging would also be merged into main. The only tests failing are the snapshots, which I generated using the wrong .env file. I just fixed it and am waiting for the action to run once again.

Sorry. Those who read have a great advantage.

ff6347 (approved):
LGTM. 🚀

dnsos (approved):
Good work!

I'm not encountering this bug anymore when viewing or creating sensors (Deploy Preview environment). Curious that the error handling changed under the radar for us 😕

PR #283, opened 2022-03-23
STADTPULS-725 ci(semver release): Adds config, packages & action

ff6347:
This should enable semantic release bot to run on a github aciton and
- add a changelog
- bump the package version
- commit all those things
- create a github release from it

PR #284, opened 2022-03-23
fix(deps): adds missing eslint plugin

ff6347:
In staging there was a drift from main. At some point during the update of nextjs this plugin
slipped away


ff6347:
fixed by #284

vogelino:
Already attempting to fix that in #283. Should have done it in a separate PR. However, it should work now.

PR #307, opened 2023-01-09
chore: remove interactions for sensor records

dnsos:
This PR disables the filtering of records. Also there is only one download button now which downloads (a maximum of) 3000 records for a sensor.


vogelino (approved):
Good job! Nice UI feedback, removal of WebWorker, etc. Let's merge it!

